PR #2, opened 2021-07-20
Development environment setup

pdyraga:
Initial development environment setup based on [keep-network/tbtc-v2](https://github.com/keep-network/tbtc-v2) and
[keep-network/coverage-pools](https://github.com/keep-network/coverage-pools/). Hardhat, eslint, solhint, prettier, slither. 
Proven weapons for our upcoming Solidity battle.

Additionally, the following CI jobs have been configured in GitHub Actions:
- contracts-build-and-test
- contracts-slither
- code-lint-and-format


package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "@t-network/solidity-contracts",
+  "version": "0.0.1",
+  "license": "MIT",
```
pdyraga, line 4: MIT, GPLv3, ...?

Shadowfiend: G-P-L, G-P-L.

IMO 😏

cygnusv: For reference, all current NuCypher contracts are AGPLv3

contracts/token/TToken.sol
```diff
@@ -0,0 +1,5 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity 0.8.4;
```
nkuba, line 3: Don't we want to use semver `^0.8.4`?

.github/workflows/contracts.yaml
```diff
@@ -0,0 +1,76 @@
...
+      - name: Install dependencies
+        run: yarn install
+
+      - name: Build contracts
+        run: yarn build
+
+      - name: Run tests
+        run: yarn test
+
+  contracts-lint:
```
nkuba, line 30: Maybe we don't need this job as `yarn lint` is a part of `yarn format` we check in the other workflow?

contracts/token/TToken.sol
cygnusv: Good eye!

pdyraga: Not using the latest release was a recommendation of one of Keep contracts auditors, Trail of Bits: 
> In terms of solc versions, 0.8.x is fine, as long it's not the latest release (which sometimes contains latent issues). Right now, latest release is 0.8.5, so 0.8.4 is fine, in particular since the fixed a high-severity issue related with the decoder.

We could use `^0.8.4` but we would have to revisit the version before the deployment. Or... we just stick with 0.8.4 which is the version we'll use for tBTC v2 token and coverage pools. What do you think?

package.json
pdyraga: What if we start with GPLv3 and then revisit so that we are not blocking the setup PR? I think AGPLv3 could affect clients having to bundle contract bindings in the source code and binaries. 🤔

.github/workflows/contracts.yaml
pdyraga: Yeah, good call. So maybe we'll drop this job and rename `code-format` job to `code-lint-and-format`? Linting is about static code analysis to flag programming errors, bugs, stylistic errors, and suspicious constructs. Formatting is just about the convention: new lines, whitespaces, semicolons, and so forth.

nkuba: 💯

cygnusv (approved):

package.json
cygnusv: Sure, let's do that 👍

pdyraga: Done in https://github.com/threshold-network/solidity-contracts/pull/2/commits/876570b7f518ec18d231e562dd5972336d00dc91.

contracts/token/TToken.sol
pdyraga: Done in 5b4a7eb2efbd39756fe24eb51291e5b3fa3cb51d but let's keep in mind to pin it down before the deployment.

nkuba (approved):

Issue #1, opened 2021-07-20
KEEP/NU <> T token transition mechanics


pdyraga:
# Overview

This issue aims at discussing RC0 Token Proposal ([[1](https://snapshot.org/#/keepstakers.eth/proposal/QmV9o1AiZSVZ5fdKJXYmGTcnYQodP8h1gWt9EEMaNFXsDs)], [[2](https://snapshot.org/#/nudao.eth/proposal/QmdH23qw5A1Afr3gT6qokn18UskpunM1cnzZpABsJtxEAv)]) implementation and how KEEP and NU holders will move their positions to T. Solutions proposed in this issue should be generic enough so that they can be easily integrated with existing contracts of Keep Network and NuCypher.

# Roles in the system

We identified 3 roles in the system essential to the migration. There are multiple variants of these roles but it is expected that understanding and implementing paths for 3 basic roles is enough and all their variants can be handled by the basic migration paths for these roles.

The roles are:
- liquid token owner,
- grantee,
- staker.

# T token

We will implement T ERC20 token based on the audited [`ERC20WithPermit`](https://github.com/keep-network/tbtc-v2/blob/main/solidity/contracts/token/ERC20WithPermit.sol) used for TBTC v2 and Coverage Pool COV token. It is a burnable ERC20 token with EIP2612 permit functionality. Users can authorize a transfer of their tokens with a signature conforming EIP712 standard instead of an on-chain transaction from their address. Anyone can submit this signature on the user's behalf by calling the permit function, as specified in EIP2612 standard, paying gas fees, and possibly performing other actions in the same transaction.

# KEEP/NU <> T migration for liquid token owners and grantees

Liquid token owners will be able to migrate their entire positions right away. Grantees will have to gradually migrate their positions by releasing vested tokens into liquid ones and migrating them the same way as liquid token owners. Keep in mind grant is a promise of tokens. One cannot wrap tokens they have promised and do not own yet. Still, they will be able to stake their KEEP/NU to earn in T network according to their existing staking policy.

We will implement a Vending Machine smart contract, similar to the one we use for [TBTC v1 <> TBTC v2 minting](https://github.com/keep-network/tbtc-v2/pull/9). There will be two instances of this contract deployed. One for KEEP<>T and another for NU<>T. The existing Vending Machine smart contract will require two modifications:
- there has to be a `mapping address → uint256` tracking the amounts minted/unminted per account so that only selected accounts can unmint, up to the amount they previously minted,
[...]

PR #3, opened 2021-07-21
KEEP/NU <> T Vending Machine

pdyraga:
Refs #1
~~Depends on #2 (keeping it as a draft until the dependency is merged)~~

This Pull Request addresses the first two points of the Plan of the game described in #1. 

`VendingMachine` contract implements a special lending protocol to enable KEEP/NU
token holders to wrap their tokens as collateral and borrow T tokens against that collateral
based on the wrap ratio. This will go on indefinitely and enable NU and KEEP token holders
to join T network without needing to buy or sell any assets. Logistically, anyone holding NU
or KEEP can wrap those assets in order to receive T. They can also unwrap T in order to go
back to the underlying asset. There will be a separate instance of this contract deployed for
KEEP holders and a separate instance of this contract deployed for NU holders.


package.json
```diff
@@ -31,5 +31,8 @@
     "solhint": "^3.3.6",
     "solhint-config-keep": "github:keep-network/solhint-config-keep",
     "typescript": "^4.3.2"
+  },
+  "dependencies": {
+    "@thesis/solidity-contracts": "github:thesis/solidity-contracts#682e515f6d21da5a1d29956f53ba47f8294ba1bb"
```
pdyraga, line 36: This will have to be changed into NPM package reference. Keeping it like that until we have an agreement on using `ERC20WithPermit`.

contracts/token/T.sol
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity 0.8.4;
+
+import "@thesis/solidity-contracts/contracts/token/ERC20WithPermit.sol";
+
+contract T is ERC20WithPermit {
```
pdyraga, line 7: Or `TToken` as originally? Just `T` reads better to me 🤷

package.json
pdyraga: Actually... it does not have to. We can keep referencing version from GitHub 🤷 

Updated to the most recent version in https://github.com/threshold-network/solidity-contracts/pull/3/commits/579d082bc04733c0095512930ee9d3b6a4432fe2.

pdyraga:
#2 got merged so I could update the code here and undraft the PR. This is ready!

contracts/vending/VendingMachine.sol
```diff
@@ -0,0 +1,169 @@
...
+        IERC20 _wrappedToken,
+        T _tToken,
+        uint256 _wrappedTokenSupply,
+        uint256 _tTokenAllocation
+    ) {
+        wrappedToken = _wrappedToken;
+        tToken = _tToken;
+        ratio =
+            (FLOATING_POINT_DIVISOR * _tTokenAllocation) /
+            _wrappedTokenSupply;
```
vzotova, line 80: What do you think about adding a few `required` for the input values? I know we control these parameters, but it costs nothing in terms of gas / code size.

contracts/vending/VendingMachine.sol
```diff
@@ -0,0 +1,169 @@
...
+    event Wrapped(
+        address indexed recipient,
+        uint256 wrappedTokenAmount,
+        uint256 tTokenAmount
+    );
+    event Unwrapped(
+        address indexed recipient,
+        uint256 tTokenAmount,
+        uint256 wrappedTokenAmount
+    );
```
vzotova, line 59: To make it perfect we can add docstrings for events too (and generate docs from contracts in some moment in a future)

vzotova (approved):

contracts/vending/VendingMachine.sol
cygnusv: Opened #11

cygnusv (approved):

PR #8, opened 2021-08-08
T token with Governance Bravo support

pdyraga:
Refs #1 

T token based on `ERC20WithPermit` from [thesis/solidity-contracts](https://github.com/thesis/solidity-contracts) with support for Governor Bravo checkpoints, so that Token Holder DAO proposed in [Threshold Network DAO Proposal v1](https://forum.threshold.network/t/threshold-network-dao-proposal-v1/26) can be implemented.

The key to understanding the logic are three functions:
- `function delegate(address delegatee) external` (or `delegateBySig`),
- `getCurrentVotes(address account) external view returns (uint96) `,
- `getPriorVotes(address account, uint256 blockNumber) external view returns (uint96)`.

Each T token holder can delegate to a selected delegatee address or self-delegate their votes. Delegatees have voting power equal to the token holder's balance. If token holder did not delegate their votes, their voting power is 0. The DAO will determine the voting power for the proposal based on the voting power one had at the moment the proposal was created. `mint`, `burn`, `burnFrom`, `transfer`, and `transferFrom` functions will automatically update one's voting power assuming the token holder delegated. Structure `Checkpoint` stores information about the voting power at the given block and a new `Checkpoint` is registered every time the voting power changes.

Most of the functionality has been ported unchanged from [COMP](https://github.com/compound-finance/compound-protocol/blob/master/contracts/Governance/Comp.sol). The most significant adjustments are:
- Updated `delegateBySig` to make the interface consistent with EIP2612 `permit` implemented in `ERC20WithPermit` super class. Specifically, renamed `expiry` to `deadline` in Delegation struct, dropped nonce from parameter list, and added signatory (in `permit`, the analogy is `owner`). I have also added validation for `v` and `s` part of the signature.
- Minting more than `(unit96).max` of T is not allowed. `ERC20WithPermit` is a generic contract and needs to support `uint256` supply and balances. Governor Bravo contracts expect that current votes and prior votes are represented by `uint96` and the Checkpoint implementation where `uint32` block number and `uint96` votes counter can be packed together is more gas friendly. It is important because a new Checkpoint could be registered for every transfer of the token. Since the initial supply of T is 10B, and we use 18 decimals for precision, `uint96` is more than enough to store token balances and keeping the votes interface working with `uint96` will let us avoid changes to Governor Bravo DAO contracts.
- Reworked the code so that Solidity 8 checked arithmetics is used to the maximum extent. 

The tests were implemented from scratch and I highly recommend starting the review from tests as it will let to better understand the expected behavior.

### DIFF
--- a/contracts/test/TestToken.sol
+++ b/contracts/test/TestToken.sol
@@ -1,0 +1,9 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+pragma solidity 0.8.4;
+
+import "@thesis/solidity-contracts/contracts/token/ERC20WithPermit.sol";
+
+contract TestToken is ERC20WithPermit {
+    constructor() ERC20WithPermit("Test Token", "TEST") {}
+}
--- a/contracts/token/T.sol
+++ b/contracts/token/T.sol
@@ -20,6 +20,10 @@ contract T is ERC20WithPermit {
     /// @notice The number of checkpoints for each account
     mapping(address => uint32) public numCheckpoints;
 
+    /// @notice The EIP-712 typehash for the delegation struct used by the contract
+    bytes32 public constant DELEGATION_TYPEHASH =
+        keccak256("Delegation(address delegatee,uint256 nonce,uint256 expiry)");
+
     /// @notice An event thats emitted when an account changes its delegate
     event DelegateChanged(
         address indexed delegator,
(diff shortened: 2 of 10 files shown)


pdyraga:
I resolved conflicts so this PR is getting for review and - if reviewers decide to approve - merge. I would like to merge this PR to `main` first, and then work on #10 separately to not tangle the branch history too much. @vzotova @nkuba @cygnusv

vzotova (approved):

cygnusv (approved):

PR #9, opened 2021-08-10
Allow only exact conversions when wrapping/unwrapping

cygnusv:
The correctness of the long-term behaviour of the Vending Machine contract will be easier to guarantee if we only allow exact conversions when wrapping/unwrapping.  Previous behavior produced unintended results when the amount converted wasn't exact wrt to the conversion ratio. E.g, for a 10:3 ratio between token A and token B, the vending machine had different balances if the token holder converts 10 A (resulting in 3 B), than converting 5 A twice (resulting in 1+1 B due to rounding). New behavior converts the max amount that can be converted exactly from the approved amount, so the caller will be left with spare change if they provide an amount that's not exact wrt to the conversion ratio.

Added also a type restriction in the supply amounts provided in the constructor, which ensures that there can't be multiplication overflow in the contract logic, as well as some additional test.

### DIFF
--- a/contracts/vending/VendingMachine.sol
+++ b/contracts/vending/VendingMachine.sol
@@ -19,7 +19,7 @@ import "../token/T.sol";
 ///         order to go back to the underlying asset. There is a separate
 ///         instance of this contract deployed for KEEP holders and a separate
 ///         instance of this contract deployed for NU holders.
-contract VendingMachine is Ownable, IReceiveApproval {
+contract VendingMachine is IReceiveApproval {
     using SafeERC20 for IERC20;
     using SafeERC20 for T;
 
--- a/contracts/vending/VendingMachine.sol
+++ b/contracts/vending/VendingMachine.sol
@@ -67,27 +67,32 @@ contract VendingMachine is IReceiveApproval {
 
     /// @notice Sets the reference to `wrappedToken` and `tToken`. Initializes
     ///         conversion `ratio` between wrapped token and T based on the
-    ///         provided `_tTokenAllocation` and `_wrappedTokenSupply`.
+    ///         provided `_tTokenAllocation` and `_wrappedTokenAllocation`.
     /// @param _wrappedToken Address to ERC20 token that will be wrapped to T
     /// @param _tToken Address of T token
-    /// @param _wrappedTokenSupply The total supply of the token that will be
+    /// @param _wrappedTokenAllocation The total supply of the token that will be
     ///       wrapped to T
     /// @param _tTokenAllocation The allocation of T this instance of Vending
     ///        Machine will receive
     /// @dev Multiplications in this contract can't overflow uint256 as we
-    ///     restrict `_wrappedTokenSupply` and `_tTokenAllocation` to 192 bits
-    ///     and the value in FLOATING_POINT_DIVISOR fits in less than 60 bits.
+    ///     restrict `_wrappedTokenAllocation` and `_tTokenAllocation` to
+    ///     192 bits and FLOATING_POINT_DIVISOR fits in less than 60 bits.
     constructor(
         IERC20 _wrappedToken,
         T _tToken,
-        uint192 _wrappedTokenSupply,
+        uint192 _wrappedTokenAllocation,
         uint192 _tTokenAllocation
     ) {
+        require(
+            _tToken.totalSupply() >= _tTokenAllocation &&
+                _wrappedToken.totalSupply() >= _wrappedTokenAllocation,
+            "Allocations can't be greater than token supplies"
+        );
         wrappedToken = _wrappedToken;
         tToken = _tToken;
         ratio =
             (FLOATING_POINT_DIVISOR * _tTokenAllocation) /
-            _wrappedTokenSupply;
+            _wrappedTokenAllocation;
     }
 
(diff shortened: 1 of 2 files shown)


contracts/vending/VendingMachine.sol
```diff
@@ -80,20 +90,20 @@ contract VendingMachine is Ownable, IReceiveApproval {
             _wrappedTokenSupply;
```
vzotova, line 90: What do you think about adding few additional checks?
* that there is no remainder of division when `ratio` is calculated (by calculating `_tTokenAllocation` back for example)
* check contracts and supplies `_tToken.totalSupply() >= _tTokenAllocation` and `_wrappedToken.totalSupply() >= _wrappedTokenSupply`

test/vending/VendingMachine.test.js
```diff
@@ -45,10 +65,24 @@ describe("VendingMachine", () => {
...
+        expect(await vendingMachine.ratio()).to.equal(expectedRatio)
+      })
+    })
+  })
+
   describe("wrap", () => {
     context("when caller has no wrapped tokens", () => {
       it("should revert", async () => {
-        const amount = 1
+        const amount = to1e18(42)
```
pdyraga, line 85: Maybe just `to1e18(1)`? It took me a while to think why 42 and I think it's just a random value, right?

test/vending/VendingMachine.test.js
```diff
@@ -242,23 +322,23 @@ describe("VendingMachine", () => {
...
           vendingMachine.connect(thirdParty).unwrap(amount)
-        ).to.be.revertedWith("Transfer amount exceeds balance")
+        ).to.be.revertedWith("Can not unwrap more than previously wrapped")
       })
     })
 
     context("when token holder has not enough T tokens", () => {
       it("should revert", async () => {
-        const amount = tAmount.add(1)
+        const amount = tAmount.add(to1e18(42))
```
pdyraga, line 335: Okay, now I started thinking there must be something special about 42 that I do not see yet... 🤔

pdyraga:
Not related to this PR but we can fix it in one shot: I realized the contract is `Ownable` while it does not need to be ownable, there is nothing `onlyOwner`.

contracts/vending/VendingMachine.sol
```diff
@@ -58,20 +65,23 @@ contract VendingMachine is Ownable, IReceiveApproval {
...
+    /// @dev Multiplications in this contract can't overflow uint256 as we
+    ///     restrict `_wrappedTokenSupply` and `_tTokenAllocation` to 192 bits
+    ///     and the value in FLOATING_POINT_DIVISOR fits in less than 60 bits.
     constructor(
         IERC20 _wrappedToken,
         T _tToken,
-        uint256 _wrappedTokenSupply,
-        uint256 _tTokenAllocation
+        uint192 _wrappedTokenSupply,
+        uint192 _tTokenAllocation
```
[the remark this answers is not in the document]
cygnusv: Restricting the supply of T to `uint96` is fine, but we just need to be aware that that allows up to 79 billion T: with an initial supply of 10 billion T, and a moderate issuance rate (less than 15%), it takes 15 years to reach that limit. So, this works, as long as inflation is not to high.

contracts/vending/VendingMachine.sol
```diff
@@ -116,32 +126,47 @@ contract VendingMachine is Ownable, IReceiveApproval {
...
+        uint256 convertibleAmount = amount - tRemainder;
+        wrappedAmount = (convertibleAmount * FLOATING_POINT_DIVISOR) / ratio;
     }
 
     function _wrap(address tokenHolder, uint256 wrappedTokenAmount) internal {
-        uint256 tTokenAmount = conversionToT(wrappedTokenAmount);
+        (uint256 tTokenAmount, uint256 remainder) = conversionToT(
+            wrappedTokenAmount
+        );
+        wrappedTokenAmount -= remainder;
```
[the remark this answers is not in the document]
cygnusv: ✅

contracts/vending/VendingMachine.sol
```diff
@@ -23,8 +23,15 @@ contract VendingMachine is Ownable, IReceiveApproval {
     using SafeERC20 for IERC20;
     using SafeERC20 for T;
 
+    /// @notice Number of decimal places of precision in conversion to/from
+    ///         wrapped tokens (assuming typical ERC20 token with 18 decimals).
+    ///         This implies that amounts of wrapped tokens below this precision
+    ///         won't take part in the conversion.
```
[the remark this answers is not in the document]
cygnusv: Added comment ✅

contracts/vending/VendingMachine.sol
cygnusv: Added the supplies check. The remainder check for ratio calculation doesn't add much, and it would simply move the remainder problem somewhere else (specifically, to the address that funds the vending machine with T)

test/vending/VendingMachine.test.js
cygnusv: Oh well, it's simply that 42 is the Answer to the Ultimate Question of Life, the Universe, and Everything, but yeah, `to1e18(1)` works fine too.

vzotova (approved):

PR #10, opened 2021-08-10
T extends MisfundRecovery, updated @thesis/solidity-contracts dependency

pdyraga:
~~Depends on #8~~
See https://github.com/thesis/solidity-contracts/pull/9
See https://github.com/thesis/solidity-contracts/pull/8

The new `@thesis/solidity-contracts` version introduces `MisfundRecovery`
contract allowing the contract owner to recover misfunded ERC721 and ERC20.

Additionally, in `ERC20WithPermit`, it prevents sending tokens to the
token contract address which should prevent most of the misfund cases
we've seen so far on mainnet for KEEP.

I have updated T token to extend `MisfundRecovery`. This way, DAO which
will be the owner of T token will be able to recover misfunds in the
future.

### DIFF
--- a/contracts/token/T.sol
+++ b/contracts/token/T.sol
@@ -3,10 +3,11 @@
 pragma solidity 0.8.4;
 
 import "@thesis/solidity-contracts/contracts/token/ERC20WithPermit.sol";
+import "@thesis/solidity-contracts/contracts/token/MisfundRecovery.sol";
 
 /// @title T token
 /// @notice Threshold Network T token.
-contract T is ERC20WithPermit {
+contract T is ERC20WithPermit, MisfundRecovery {
     /// @notice A checkpoint for marking number of votes from a given block.
     struct Checkpoint {
         uint32 fromBlock;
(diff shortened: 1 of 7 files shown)


lukasz-zimnoch (approved):

contracts/token/T.sol
```diff
@@ -3,10 +3,11 @@
 pragma solidity 0.8.4;
 
 import "@thesis/solidity-contracts/contracts/token/ERC20WithPermit.sol";
+import "@thesis/solidity-contracts/contracts/token/MisfundRecovery.sol";
 
 /// @title T token
 /// @notice Threshold Network T token.
-contract T is ERC20WithPermit {
+contract T is ERC20WithPermit, MisfundRecovery {
```
nkuba, line 10: Can we add a simple happy-path test case to check that misfund recovery is possible for `T` contract? I know it's an external contract tested elsewhere but we should check that `T` contract gives a basic functionality expected after extending the `MisfundRecovery` contract.

nkuba (changes requested):

nkuba (approved):

vzotova (approved):

Issue #14, opened 2021-08-20
"Liquidation" mechanism to guarantee synchronization between T allocations and KEEP legacy staking contract


cygnusv:
During last call (@pdyraga @nkuba @vzotova and myself) we talked about the synchronization issues between KEEP legacy staking contract and T staking, which is specially tricky as KEEP staking contract is not upgradeable. The expected order of operations for a legacy KEEP staker that wants to unstake should be:
1) De-allocate from T applications
2) Unstake from KEEP contract

Otherwise, if they unstake first, then T applications would have inconsistent information. As a deterrent, we will allow anyone that detects this situation to liquidate this stake by seizing it (and getting the 5% fee allowed by KEEP staking contract). Note that this won't be necessary for legacy NU stakers and native T stakers since we can encode these restrictions on the NU and T side, respectively.

cygnusv:
Closed by #7

Issue #13, opened 2021-08-20
De-allocation logic


cygnusv:
@vzotova and I have been thinking about the specifics of the staking contract with respect to application allocations, and we have some questions. First, let's settle on some terminology for the moment:

- **Allocate**: Assign an amount of $T that belongs to a T stake (or equivalent legacy stake in $NU/$KEEP) to a T application. There's no allocation overlapping (i.e., you can't reuse the same portion of stake to different applications)
- **De-allocate**: Taking back an allocation from an application. There are 2 types of de-allocation:
    - **Cancel allocation**: The allocated amount that's being taken back is not allocated elsewhere
    - **Switch allocation**: The allocated amount is moved to a different application

We think that there should be different rules governing canceling and switching allocations: there are different intentions in someone moving $T from PRE to tBTC apps than from someone simply wanting to withdraw. In other words, switching allocations should be less restrictive than cancelling allocations. Also, we think that this rules would probably be different _per application_ (e.g., it may be different to de-allocate from PRE than de-allocating from tBTC), for instance, in terms of the effective time to de-allocate.

What do you think about this? With respect to tBTCv2, how does this look like? Do you see different rules about cancelling and switching allocations from tBTC to another T application?

Another related question is about _partial de-allocations_, in terms of amount of tokens being de-allocated. We don't see any different from full de-allocations from the perspective of the PRE application, but what about tBTCv2? Whats the process like for a staker wanting to, say, de-allocate half of their tBTCv2 stake allocation?

pdyraga:
Are we good about closing this issue now @cygnusv?

PR #15, opened 2021-09-07
RFC 1: T staking contract

pdyraga:
Refs #1 
Refs #7 

The goal of this proposal is to specify a simple and secure stake delegation
mechanism. It should enable T owners to have their wallets offline and their
stake operated by operators on their behalf. All off-chain client software
should be able to run without exposing operator’s private key and should not
require any owner’s keys at all. The stake delegation should also optimize the
network throughput without compromising the security of the owners’ stake.

This proposal aims at implementing a minimum viable staking contract version
allowing to support legacy KEEP and NU stakes, as well as native T delegations
with a minimum effort. It proposes a naming and full public API of the contract
we can discuss and agree on before undertaking any further implementation
effort.

The functionality of the staking contract can be further extended by the
upgradeability of the contract code.

### DIFF
--- a/docs/rfc-1-staking-contract.adoc
+++ b/docs/rfc-1-staking-contract.adoc
@@ -61,27 +61,23 @@ Tokens are delegated by the owner. During the delegation, the owner needs to
 appoint an operator, beneficiary, and authorizer. The owner may decide to
 delegate just a portion of their tokens. The owner may delegate multiple times
 to different operators. For simplicity, the operator address can not be reused
-between delegations even if the delegation has been recovered. The only
-exception is when the delegation has been canceled. To protect against griefing
-by frontrunning stake delegation transactions and reserving operator addresses,
-staking contract should require a governable minimum stake amount for each
-delegation.
+between delegations even if the tokens have been undelegated. To protect against 
+griefing by frontrunning stake delegation transactions and reserving operator
+addresses, staking contract should require a governable minimum stake amount for 
+each delegation and there should be a minimum undelegation delay of 24 hours.
 
 Staking contract binds the owner with the operator, beneficiary, and authorizer
 for the given stake delegation and records the stake delegation amount.
-Delegation is set to active after 2 hours, once the cancellation window passes.
 
 Existing KEEP and NU stake owners also need to perform a delegation so that T
 staking contract can record their operator, beneficiary, and authorizer. Since
 these values are read-only, they should be copied from KEEP staking contract to
 T staking contract and NU stakers should set them during the delegation. For
-existing stakes, the entire stake amount is registered in T staking contract.
-
-There are 2 hours to cancel a stake without suffering an undelegation period.
-This functionality is needed e.g. in case of a typo in the operator address.
-Undelegation can not be canceled in the same transaction in which it was
-created. Once the delegation has been canceled, the operator address can be used
-again.
+existing stakes, the entire amount staked in the legacy contract is cached in T
+staking contract. Caching amount staked in the legacy staking contract helps to 
+identify discrepancy between staking contracts. It is much easier to just compare
+two values instead of checking authorization of every application (see Keeping
+information in sync section).
 
 ==== Authorizing an application
 
@@ -92,24 +88,41 @@ is authorized to slash or seize the stake.
 
 The authorizer decides up to which amount the given application is eligible to
 slash. This allows managing the risk profile of the staker individually for each
-application. The authorized stake amount can only be increased.
+application. The authorized stake amount can be increased and decreased.
 
-Authorization can not be canceled.
+The amount authorized is set individually for each application and should be
+used as a staker's weight in the application: for work selection and rewards.
+The amount authorized for the application can be also decreased as a result of
+slashing, including slashing executed in the context of another application. 
 
 The application being authorized is notified via a callback every single time
 new authorization happens and the application may decide to revert the
-transaction if individual requirements of the application are not met. Such
-requirements could be, for example, not enough stake authorized for the
-particular application or not supported legacy stake (see Limitations section).
+transaction if individual requirements of the application are not met.
+The application is notified every single time the authorization increases or
+decreases. The application can revert this transaction in case some rules
+specific to that application are not met. Inside the callback, the application
+can also update its internal state, such as, for example, the weight of the operator
+in the sortition pool.
+
+Authorization decrease is a two-step process. First, the authorizer requests to
+decrease the authorization, and the application is notified about it by
+a callback. In the second step, the application approves the decrease at its own
+discretion. It is up to every individual application to decide if and when the
+authorization decrease gets approved. New authorization decrease request
+overwrites the pending one.
 
 All applications first need to be authorized in a registry of T applications
-controlled by the governance.
+controlled by the governance before any authorizer can use them.
+To protect against the attack of blocking the slashing queue by exceeding the
+block gas limit for slashing the particular operator, there should be a
+governable limit of applications that can be authorized per operator.
 
 ==== Stake top-up
 
-Top-up increases the amount of tokens locked in a delegation. This increases the
-probability of being chosen for work in the network but is only effective for
-future checks of the stake amount.
+Top-up increases the amount of tokens locked in a delegation and allows to later
+increase the authorization for applications. This increases the probability of being
+chosen for work in the network but is only effective for future checks of the
+authorized amount.
 
 Top-ups can be executed for native T stakes and for legacy KEEP/NU stakes.
 
@@ -132,47 +145,44 @@ increasing their rewards.
 
 Anyone can execute a stake top-up for an operator using a liquid T. Stake top-up
 does not automatically increase authorization levels for applications.
+Stake top-up is a one-step process and does not require any delay.
 
 ==== Undelegating a stake
 
-The owner or operator may choose to undelegate the stake by submitting an
-undeleagation order. Stake delegation should be marked in the staking contract
-as undelegating. Always the entire stake is getting undelegated and there is a
-fixed, non-governable, two months undelegation period. All applications should
-be designed in such a way that a two-month undelegation period is sufficient to
-retire all groups undelegating operators are involved in. The operator is not
-eligible for any new work during the stake undelegation period but it is still
-required to finish all the pending work. Authorized applications are eligible
-to slash the operator for the entire undelegation period.
+The owner or operator may choose to undelegate the stake if there are no
+authorizations on applications or if all authorizations have been decreased to
+zero. It is not possible to partially undelegate the stake. Undelegation returns
+all tokens to the owner but does not clear information about the delegation so
+that any application can retrieve this information later to, for example, pay
+out rewards.
 
-==== Recovering a stake
+If the undelegation happens before 24 hours passed since the delegation, the
+transaction reverts.
 
-Once the undelegation period finishes, anyone - even a third party - can recover
-the stake. Staking contract sets the stake as recovered and returns all
-delegated tokens to the owner. Applications shouldn't be able to slash after
-undelegation is complete, even before the stake is returned. Information about
-owner, operator, and beneficiary, is retained in the contract in case some
-application rewards are still available for withdrawal.
+It is expected that undelegation is first completed on T staking contract before
+undelegation on a legacy staking contract for the given operator gets initiated.
 
 ==== Keeping information in sync
 
-To avoid expensive calls to legacy staking contracts, it is assumed that
-information in T staking contract about the staked amount is always up-to-date.
-By design, this is always the case for native T stakers but might not be the
-case for legacy KEEP/NU stakers.
+To avoid expensive calls to legacy staking contract, it is assumed that cached
+information in T staking contract about the amount staked in the legacy contract
+is always up-to-date.
 
 T staking contract should expose a function allowing to seize some amount of T
-from the operator whose stake in T contract is lower than their stake in the
-legacy staking contract. 5% of the amount seized is given to the person who
+from the operator in case that operator has a lower active stake
+(eligible for work selection) in the old staking contract than the amount cached
+in T staking contract. 5% of the amount seized is given to the person who
 notified about the discrepancy and the rest is burned. The amount is a
 governable parameter and can be updated at any time by the governance, with no
-governance delay.
+governance delay. The transaction notifying about stake amounts not being in sync
+needs to update authorizations of all affected applications and execute an
+involuntary allocation decrease on each affected application.
 
 For legacy stakers, staked amount can become out-of-sync in three cases:
 
 * stake undelegated on the legacy contract,
-* stake slashed on the legacy contract,
-* stake topped-up on the legacy contract.
+* stake topped-up on the legacy contract,
+* stake slashed on the legacy contract.
 
 It is expected that stake undelegation will be first performed on T staking
 contract and then on the legacy staking contract, in the same transaction.
@@ -185,20 +195,56 @@ is higher than the stake amount on T staking contract.
 
 In case the stake has been slashed on the legacy contract, the operator is
 required to update their information on T staking contract as soon as possible.
-In practice, with the random beacon disabled, and tBTC v1 slashing the stake
-only in case of a proven fraud that had to be committed by all operators of
+In practice, for Keep, with the random beacon disabled, and tBTC v1 slashing the 
+stake only in case of a proven fraud that had to be committed by all operators of
 ECDSA keep, this approach is acceptable.
 
-An integral part of the staking contract should be a bot monitoring stakes and
-notifying about discrepancies. This is especially important given that the bot
-may need to voluntarily inform about discrepancies for operators that have been
-slashed to zero.
+An integral part of the staking contract should be a bot or process inside an
+off-chain client monitoring stakes and notifying about discrepancies. This is
+especially important given that the bot may need to voluntarily inform about
+discrepancies for operators that have been slashed to zero.
+
+Owner or operator can decrease the legacy contract active stake cached amount on
+T staking contract if no application has authorization higher than the liquid T
+stake. It allows to undelegate from the legacy staking contract while still
+being able to operate in T and earn rewards.
 
 ==== Slashing a stake
 
 Authorized applications can slash or seize a stake. Slash operation decreases
 the stake of an operator and burns slashed tokens. Seize decreases the stake,
-burns 95% of the stake, and awards 5% to the notifier of misbehavior. 
+burns 95% of the stake, and awards up to 5% to the notifier of misbehavior.
+
+To keep stakes synchronized between applications when operators are slashed,
+without the risk of running out of gas, the staking contract queues up slashings
+and let users process the transactions.
+
+When an application slashes one or more operators, it adds them to the slashing
+queue on the staking contract. A queue entry contains the operator's address and
+the amount they're due to be slashed.
+
+When there is at least one operator in the slashing queue, any account can
+submit a transaction processing one or more operators' slashings, and collecting
+a reward for doing so. A queued slashing is processed by updating the operator's
+stake to the post-slashing amount, updating authorized amount for each
+affected application, and notifying all affected applications that
+the operator's authorized stake has been reduced due to slashing. The
+application must then do the necessary adjustments, such as removing the
+operator from the sortition pool or reducing its weight, changing the operator's
+eligibility for rewards, and so forth.
+
+Every application callback executed as a result of slash should have a 250k gas
+limit. Slashing are processed in a FIFO basis, and there is just one function
+exposed by the staking contract allowing to slash one or more operators from the
+head of the queue. Callback failure does not revert the transaction.
+
+In the case of legacy stakers, their liquid T is slashed first before a call to 
+the legacy contract is executed.
+
+It is important to note slashing executed in the context of one application may
+lead to involuntarily decreasing the authorization for other applications in 
+case the amount of stake available after the slashing is lower than these
+authorizations.
 
 === Limitations
 
@@ -228,13 +274,15 @@ transfer to the staking contract.
 ==== `stakeKeep(address operator) external`
 
 Copies delegation from the legacy KEEP staking contract to T staking contract.
-No tokens are transferred. Can be called by anyone.
+No tokens are transferred. Caches the active stake amount from KEEP staking 
+contract. Can be called by anyone.
 
 ==== `stakeNu(address operator, address beneficiary, address authorizer) external`
 
 Copies delegation from the legacy NU staking contract to T staking contract,
-additionally appointing beneficiary and authorizer roles. Can be called only by
-the original delegation owner.
+additionally appointing beneficiary and authorizer roles. Caches the amount
+staked in NU staking contract. Can be called only by the original delegation
+owner.
 
 ==== `setMinimumStakeAmount(uint256 amount) external onlyGovernance`
 
@@ -255,18 +303,44 @@ Authorizes the particular application for the given operator. From this moment,
 the application may slash operator's stake up to the given amount. Can only be
 called by the given operator's authorizer.
 
-==== `disableApplication(address application) external onlyPanicButton`
+==== `increaseAuthorization(address operator, address application, uint256 amount) external onlyAuthorizerOf(operator)`
+
+Increases the authorization of the given operator for the given application by
+the given amount. Calls `authorizationIncreased(address operator, uint256 amount)`
+callback on the given application to notify the application. Can only be called
+by the given operator's authorizer.
+
+==== `requestAuthorizationDecrease(address operator, address application, uint256 amount) external onlyAuthorizerOf(operator)`
+
+Requests decrease of the authorization for the given operator on the given
+application by the provided amount. Calls `authorizationDecreaseRequested(address operator, uint256 amount)`
+on the application. It does not change the authorized amount. Can only be called
+by the given operator's authorizer. Overwrites pending authorization decrease
+for the given operator and application.
+
+==== `approveAuthorizationDecrease(address operator) external onlyRequestedApplication`
+
+Called by the application at its discretion to approve the previously requested
+authorization decrease request. Can only be called by the application that
+was previously requested to decrease the authorization for that operator.
+
+==== `disableApplication(address application) external onlyPanicButtonOf(application)`
 
 Disables the given application's eligibility to slash stakes. Can be called only
 by a panic button of the particular application. The disabled application can not
-slash stakes until it is approved again by the governance. Should be used only
-in case of an emergency.
+slash stakes until it is approved again by the governance using `approveApplication`
+function. Should be used only in case of an emergency.
 
 ==== `setPanicButton(address application, address panicButton) external onlyGovernance`
 
 Sets the panic button role for the given application to the provided address.
 Can only be called by the governance. If the panic button for the given
 application should be disabled, the role address should can set to 0x0 address.
+
+==== `setAuthorizationCeiling(uint256 ceiling) external onlyGovernance`
+
+Sets the maximum number of applications one operator can authorize. Used to
+protect against DoSing slashing queue. Can only be called by the governance.
 
 === Stake top-up
 
@@ -288,40 +362,53 @@ to T staking contract. Can be called by anyone.
 
 === Undelegating a stake
 
-==== `cancelStake(address operator) external`
-
-Cancels T stake delegation. Can not be called in the same transaction as the
-original delegation order. Reverts after 2 hours from the time the delegation
-order was created. Clears up information about the operator so that the
-operator's address can be reused. Can be called by owner and operator.
-
 ==== `undelegate(address operator) external`
 
-Orders undelegation from the given operator. Sets the delegation state as
-undelegating. Can be called only by the owner or operator.
+Performs undelegation from the given operator returning tokens to the owner. 
+Can be called only by the owner or operator. Reverts if called earlier than 24h
+after the delegation has been created. Reverts if there is at least one non-zero
+application authorization for the given operator.
 
-=== Recovering a stake
+==== `unstakeKeep(address operator) external`
 
-==== `recoverStake(address operator) external`
+Sets the legacy staking contract active stake amount cached in T staking
+contract to 0. Reverts if the amount of liquid T staked in T staking contract is
+lower than the highest application authorization. This function allows to
+unstake from Keep staking contract and sill being able to operate in T network
+and earning rewards based on the liquid T staked. Can be called only by the
+delegation owner and operator.
 
-Sets the undelegating delegation state as recovered once the undelegation period
-passes. For liquid T delegations, transfers tokens back to the original owner.
-Can be called by anyone.
+==== `unstakeNu(address operator) external`
+
+Sets the legacy staking contract active stake amount cached in T staking
+contract to 0. Reverts if the amount of liquid T staked in T staking contract is
+lower than the highest application authorization. This function allows to
+unstake from Nu staking contract and sill being able to operate in T network
+and earning rewards based on the liquid T staked. Can be called only by the
+delegation owner and operator.
 
 === Keeping information in sync
 
 ==== `notifyKeepStakeDiscrepancy(address operator)`
 
-Notifies about the discrepancy between legacy KEEP stake and stake amount in T
-staking contract. Can be called by anyone, notifier receives a reward.
+Notifies about the discrepancy between legacy KEEP active stake and amount
+cached in T staking contract. Slashes the operator in case the amount cached
+is higher than the actual active stake amount in KEEP staking contract. 
+Needs to update authorizations of all affected applications and execute an
+involuntary allocation decrease on all affected applications.
+Can be called by anyone, notifier receives a reward.
 
 Optionally: reward withdrawal can be split into a separate function to protect
 against MEV frontrunners. 
 
 ==== `notifyNuStakeDiscrepancy(address operator)`
 
-Notifies about the discrepancy between legacy NU stake and stake amount in T
-staking contract. Can be called by anyone, notifier receives a reward.
+Notifies about the discrepancy between legacy NU active stake and amount
+cached in T staking contract. Slashes the operator in case the amount cached
+is higher than the actual active stake amount in NU staking contract.
+Needs to update authorizations of all affected applications and execute an
+involuntary allocation decrease on all affected applications.
+Can be called by anyone, notifier receives a reward.
 
 Optionally: reward withdrawal can be split into a separate function to protect
 against MEV frontrunners. 
@@ -337,45 +424,56 @@ tokens are burned. Can only be called by the governance. See `seize` function.
 
 ==== `slash(uint256 amount, address[] memory operators) external onlyAuthorizedApplication`
 
-Slashes the provided amount from the stake of every operator in the array.
-Can only be called by an authorized application.
+Adds operators to the slashing queue along with the amount that should be
+slashed from each one of them. Can only be called by an authorized application.
 
 ==== `seize(uint256 amount, uint256 rewardMultipier, address notifier, address[] memory operators) external onlyAuthorizedApplication`
 
-Seize the provided amount from the stake of every operator in the array.
-The misbehavior notifier is rewarded with 5% of the total seized amount scaled
-by the reward adjustment parameter. The rest of the tokens are burned. Can only
-be called by an authorized application.
+Adds operators to the slashing queue along with the amount, reward multiplier
+and notifier address. The notifier will receive 1% of the slashed amount scaled
+by the reward adjustment parameter once the seize order will be processed. Can
+only be called by an authorized application.
+
+==== `processSlashing(uint256 count)`
+
+Takes the `count` of queued slashing operations and processes them. Receives 5%
+of the slashed amount if the slashing request was created by the application with
+a `slash` call and 4% of the slashed amount if the slashing request was created
+by the application with `seize` call. Executes `involuntaryAllocationDecrease`
+function on each affected application.
 
 === Auxiliary functions
 
-==== `eligibleStake(address operator, address application) external view returns (uint256)`
+==== `authorizedStake(address operator, address application) external view returns (uint256)`
 
-Returns the eligible stake amount of the operator for the application.
-An eligible stake is a stake that is not currently undelegating and the given
-application had to be approved for use by the operator's authorizer. If the
-stake is undelegation or the application was not approved, returns 0.
-
-The difference between eligible and active stake is that the active stake does
-not make the operator eligible for new work selection but it may still finish
-earlier work it is assigned to before the stake is recovered.
-
-==== `activeStake(address operator, address application) external view returns (uint256)`
-
-Returns the active stake amount of the operator for the application. An active
-stake is a stake that may be in the process of undelegation but has not been
-recovered yet. The given application had to be approved for use by the
-operator's authorizer. If the stake does not exist or if it has been recovered,
-returns 0.
-
-The difference between eligible and active stake is that the active stake does
-not make the operator eligible for new work selection but it may still finish
-earlier work it is assigned to before the stake is recovered.
-
+Returns the authorized stake amount of the operator for the application.
 
 ==== `hasStakeDelegated(address operator) external view returns (bool)`
 
 Checks if the specified operator has a stake delegated and if it has been
-authorized for at least one application. The stake may be in the undelegation
-process. If this function returns true, off-chain client of the given operator
-is eligible to join the network.
+authorized for at least one application. If this function returns true,
+off-chain client of the given operator is eligible to join the network.
+
+=== Application callbacks
+
+==== `authorizationIncreased(address operator, uint256 amount)`
+
+Used by T staking contract to inform the application the the authorized amount
+for the given operator increased. The application may do any necessary housekeeping
+necessary.
+
+==== `authorizationDecreaseRequested(address operator, uint256 amount)`
+
+Used by T staking contract to inform the application that the given operator
+requested to decrease the authorization to the given amount. The application
+should mark the authorization as pending decrease and respond to the staking
+contract with `approveAuthorizationDecrease` at its discretion. Note it may
+happen right away but it also may happen several months later.
+
+==== `involuntaryAllocationDecrease(address operator, uint256 amount)`
+
+Used by T staking contract to inform the application the authorization has
+been decreased for the given operator to the given amount involuntarily, as
+a result of slashing. Lets the application to do any housekeeping neccessary.
+Called with 250k gas limit and does not revert the transaction if 
+`involuntaryAllocationDecrease` call failed.
--- a/docs/rfc-1-staking-contract.adoc
+++ b/docs/rfc-1-staking-contract.adoc
@@ -395,7 +395,7 @@ on the legacy staking contract. This function allows to unstake from NU staking
 contract and sill being able to operate in T network and earning rewards based
 on the liquid T staked. Can be called only by the delegation owner and operator.
 
-=== `unstakeAll(address operator)`
+==== `unstakeAll(address operator)`
 
 Sets cached legacy stake amount to 0, sets the liquid T stake amount to 0 and
 withdraws all liquid T from the stake to the owner. Reverts if there is at least one
@@ -473,7 +473,7 @@ off-chain client of the given operator is eligible to join the network.
 ==== `authorizationIncreased(address operator, uint256 amount)`
 
 Used by T staking contract to inform the application the the authorized amount
-for the given operator increased. The application may do any necessary housekeeping
+for the given operator increased. The application may do any housekeeping
 necessary.
 
 ==== `authorizationDecreaseRequested(address operator, uint256 amount)`
@@ -481,7 +481,7 @@ necessary.
 Used by T staking contract to inform the application that the given operator
 requested to decrease the authorization to the given amount. The application
 should mark the authorization as pending decrease and respond to the staking
-contract with `approveAuthorizationDecrease` at its discretion. Note it may
+contract with `approveAuthorizationDecrease` at its discretion. It may
 happen right away but it also may happen several months later.
 
 ==== `involuntaryAllocationDecrease(address operator, uint256 amount)`
--- a/docs/rfc-1-staking-contract.adoc
+++ b/docs/rfc-1-staking-contract.adoc
@@ -1,0 +1,381 @@
+:toc: macro
+
+= RFC 1: T Staking Contract
+
+:icons: font
+:numbered:
+toc::[]
+
+== Proposal
+
+=== Goal
+
+The goal of this proposal is to specify a simple and secure stake delegation
+mechanism. It should enable T owners to have their wallets offline and their
+stake operated by operators on their behalf. All off-chain client software
+should be able to run without exposing operator’s private key and should not
+require any owner’s keys at all. The stake delegation should also optimize the
+network throughput without compromising the security of the owners’ stake.
+
+This proposal aims at implementing a minimum viable staking contract version
+allowing to support legacy KEEP and NU stakes, as well as native T delegations.
+The functionality of the staking contract can be further extended by the
+upgradeability of the contract code.
+
+=== Terminology
+
+Owner:: An address owning T tokens and tokens that could be converted to
+T: KEEP, NU, including grants. An owner is the ultimate holder of the tokens.
+Before stake delegation, the owner has full control over the tokens, and the
+tokens are returned to the owner after stake delegation has finished.
+Owner’s participation is not required in the day-to-day operations on the
+stake, so cold storage can be accommodated to the maximum extent.
+
+Operator:: An address of a party authorized to operate in the network on behalf
+of a given owner. The operator handles the everyday operations on the delegated
+stake without actually owning the staked tokens. An operator can not simply
+transfer away delegated tokens, however, it should be noted that the operator’s
+misbehavior may result in slashing tokens and thus the entire staked amount is
+indeed at stake.
+
+Beneficiary:: An address where the rewards for participation are sent, earned by
+the operator, on behalf of the owner. A beneficiary doesn’t sign or publish any
+protocol-relevant transactions, but any currency or tokens earned by the
+operator will be transferred to the beneficiary.
+
+Authorizer:: An address appointed by an owner to authorize applications on
+behalf of them. An application must be approved by the authorizer before the
+operator is eligible to participate.
+
+Delegated stake:: An owner’s staked tokens, delegated to the operator by the
+owner. Delegation enables KEEP owners to have their wallets offline and their
+stake operated by operators on their behalf.
+
+== Specification
+
+=== Functionality
+
+==== Delegating a stake
+
+Tokens are delegated by the owner. During the delegation, the owner needs to
+appoint an operator, beneficiary, and authorizer. The owner may decide to
+delegate just a portion of their tokens. The owner may delegate multiple times
+to different operators. For simplicity, the operator address can not be reused
+between delegations even if the delegation has been recovered. The only
+exception is when the delegation has been canceled. To protect against griefing
+by frontrunning stake delegation transactions and reserving operator addresses,
+staking contract should require a governable minimum stake amount for each
+delegation.
+
+Staking contract binds the owner with the operator, beneficiary, and authorizer
+for the given stake delegation and records the stake delegation amount.
+Delegation is set to active after 2 hours, once the cancellation window passes.
+
+Existing KEEP and NU stake owners also need to perform a delegation so that T
+staking contract can record their operator, beneficiary, and authorizer. Since
+these values are read-only, they should be copied from KEEP staking contract to
+T staking contract and NU stakers should set them during the delegation. For
+existing stakes, the entire stake amount is registered in T staking contract.
+
+There are 2 hours to cancel a stake without suffering an undelegation period.
+This functionality is needed e.g. in case of a typo in the operator address.
+Undelegation can not be canceled in the same transaction in which it was
+created. Once the delegation has been canceled, the operator address can be used
+again.
+
+==== Authorizing an application
+
+Before the operator is eligible to participate in the given application, the
+authorizer appointed during the stake delegation needs to review the application
+and approve it to use the stake. From the moment of approval, the application
+is authorized to slash or seize the stake.
+
+The authorizer decides up to which amount the given application is eligible to
+slash. This allows managing the risk profile of the staker individually for each
+application. The authorized stake amount can only be increased.
+
+Authorization can not be canceled.
+
+The application being authorized is notified via a callback every single time
+new authorization happens and the application may decide to revert the
+transaction if individual requirements of the application are not met. Such
+requirements could be, for example, not enough stake authorized for the
+particular application or not supported legacy stake (see Limitations section).
+
+All applications first need to be authorized in a registry of T applications
+controlled by the governance.
+
+==== Stake top-up
+
+Top-up increases the amount of tokens locked in a delegation. This increases the
+probability of being chosen for work in the network but is only effective for
+future checks of the stake amount.
+
+Top-ups can be executed for native T stakes and for legacy KEEP/NU stakes.
+
+Native T stakers can only top-up their stakes with a liquid T.
+
+Existing KEEP and NU stakers wanting to execute a top-up have two options. One
+option is to wrap their KEEP/NU to T and then, execute a top-up in T staking
+contract. The second option is to execute a top-up in their legacy staking
+contracts and notify T staking contract about the fact their legacy stake
+increased.
+
+Effectively, it means that existing KEEP stakers can mix their legacy KEEP
+stakes with liquid T stakes. Similarly, existing NU stakers can mix their legacy
+NU stakes with liquid T stakes. This functionality adds some complexity to the
+staking contract but it puts existing KEEP/NU stakers in the same spot as new T
+stakers. Without it, existing stakers would not be able to top-up their stakes
+with T earned from operating in the network, so they would be in a worse spot
+than new T stakers allowed to top-up their stakes using earned T and this way
+increasing their rewards.
+
+Anyone can execute a stake top-up for an operator using a liquid T. Stake top-up
+does not automatically increase authorization levels for applications.
+
+==== Undelegating a stake
+
+The owner or operator may choose to undelegate the stake by submitting an
+undeleagation order. Stake delegation should be marked in the staking contract
+as undelegating. Always the entire stake is getting undelegated and there is a
+fixed, non-governable, two months undelegation period. All applications should
+be designed in such a way that a two-month undelegation period is sufficient to
+retire all groups undelegating operators are involved in. The operator is not
+eligible for any new work during the stake undelegation period but it is still
+required to finish all the pending work. Authorized applications are eligible
+to slash the operator for the entire undelegation period.
+
+==== Recovering a stake
+
+Once the undelegation period finishes, anyone - even a third party - can recover
+the stake. Staking contract sets the stake as recovered and returns all
+delegated tokens to the owner. Applications shouldn't be able to slash after
+undelegation is complete, even before the stake is returned. Information about
+owner, operator, and beneficiary, is retained in the contract in case some
+application rewards are still available for withdrawal.
+
+==== Keeping information in sync
+
+To avoid expensive calls to legacy staking contracts, it is assumed that
+information in T staking contract about the staked amount is always up-to-date.
+By design, this is always the case for native T stakers but might not be the
+case for legacy KEEP/NU stakers.
+
+T staking contract should expose a function allowing to seize some amount of T
+from the operator whose stake in T contract is lower than their stake in the
+legacy staking contract. 5% of the amount seized is given to the person who
+notified about the discrepancy and the rest is burned. The amount is a
+governable parameter and can be updated at any time by the governance, with no
+governance delay.
+
+For legacy stakers, staked amount can become out-of-sync in three cases:
+
+* stake undelegated on the legacy contract,
+* stake slashed on the legacy contract,
+* stake topped-up on the legacy contract.
+
+It is expected that stake undelegation will be first performed on T staking
+contract and then on the legacy staking contract, in the same transaction.
+
+It is expected that a top-up will be first performed on the legacy staking
+contract, and then propagated to the new staking contract, in the same
+transaction. Even if it does not happen in the same transaction, this kind of
+discrepancy is not slashable given that the stake amount on the legacy contract
+is higher than the stake amount on T staking contract.
+
+In case the stake has been slashed on the legacy contract, the operator is
+required to update their information on T staking contract as soon as possible.
+In practice, with the random beacon disabled, and tBTC v1 slashing the stake
+only in case of a proven fraud that had to be committed by all operators of
+ECDSA keep, this approach is acceptable.
+
+An integral part of the staking contract should be a bot monitoring stakes and
+notifying about discrepancies. This is especially important given that the bot
+may need to voluntarily inform about discrepancies for operators that have been
+slashed to zero.
+
+==== Slashing a stake
+
+Authorized applications can slash or seize a stake. Slash operation decreases
+the stake of an operator and burns slashed tokens. Seize decreases the stake,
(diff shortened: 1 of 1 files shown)


docs/rfc-1-staking-contract.adoc
```diff
@@ -0,0 +1,381 @@
...
+is authorized to slash or seize the stake.
+
+The authorizer decides up to which amount the given application is eligible to
+slash. This allows managing the risk profile of the staker individually for each
+application. The authorized stake amount can only be increased.
+
+Authorization can not be canceled.
+
+The application being authorized is notified via a callback every single time
+new authorization happens and the application may decide to revert the
```
eth-r, line 100: We should clarify whether this applies to just the first auhorisation or if it also applies when the authorised amount is increased; I would prefer the latter.

docs/rfc-1-staking-contract.adoc
```diff
@@ -0,0 +1,381 @@
...
+particular application or not supported legacy stake (see Limitations section).
+
+All applications first need to be authorized in a registry of T applications
+controlled by the governance.
+
+==== Stake top-up
+
+Top-up increases the amount of tokens locked in a delegation. This increases the
+probability of being chosen for work in the network but is only effective for
+future checks of the stake amount.
```
vzotova, line 112: Looks like `This increases the probability of being chosen for work in the network but is only effective for
future checks of the stake amount.` conflicts with `Stake top-up does not automatically increase authorization levels for applications.`

docs/rfc-1-staking-contract.adoc
```diff
@@ -0,0 +1,381 @@
...
+==== Authorizing an application
+
+Before the operator is eligible to participate in the given application, the
+authorizer appointed during the stake delegation needs to review the application
+and approve it to use the stake. From the moment of approval, the application
+is authorized to slash or seize the stake.
+
+The authorizer decides up to which amount the given application is eligible to
+slash. This allows managing the risk profile of the staker individually for each
+application. The authorized stake amount can only be increased.
```
eth-r, line 95: Because we notify the application, it would actually be possible to let the authorised stake be reduced if the application approves. This would require some complexity but it would be entirely on the application-side and optional (for the basic behaviour we can just prohibit stake reductions in the application). Probably not worth the complexity though.

docs/rfc-1-staking-contract.adoc
```diff
@@ -0,0 +1,381 @@
...
+
+The owner or operator may choose to undelegate the stake by submitting an
+undeleagation order. Stake delegation should be marked in the staking contract
+as undelegating. Always the entire stake is getting undelegated and there is a
+fixed, non-governable, two months undelegation period. All applications should
+be designed in such a way that a two-month undelegation period is sufficient to
+retire all groups undelegating operators are involved in. The operator is not
+eligible for any new work during the stake undelegation period but it is still
+required to finish all the pending work. Authorized applications are eligible
+to slash the operator for the entire undelegation period.
```
vzotova, line 146: Earlier we discuss that this period will be set per application basis and as result deallocation process also will be per application. After this T tokens can be withdrawable. Do we want to have second layer of undelegation?

docs/rfc-1-staking-contract.adoc
```diff
@@ -0,0 +1,381 @@
...
+   
+Creates a delegation with `msg.sender` owner with the given operator,
+beneficiary, and authorizer. Transfers the given amount of T to the staking
+contract. The owner of the delegation needs to have the amount approved to
+transfer to the staking contract.
+
+==== `stakeKeep(address operator) external`
+
+Copies delegation from the legacy KEEP staking contract to T staking contract.
+No tokens are transferred. Can be called by anyone.
```
eth-r, line 231: We don't need to let anyone call this, even though we can't determine the ultimate owner of a KEEP stake easily; the operator's authoriser needs to approve the T staking contract on the KEEP staking contract (otherwise T couldn't slash), and this is exactly what the authoriser role exists for.

Thus we can limit `stakeKeep()` to "operator only".

docs/rfc-1-staking-contract.adoc
```diff
@@ -0,0 +1,381 @@
...
+==== Keeping information in sync
+
+To avoid expensive calls to legacy staking contracts, it is assumed that
+information in T staking contract about the staked amount is always up-to-date.
+By design, this is always the case for native T stakers but might not be the
+case for legacy KEEP/NU stakers.
+
+T staking contract should expose a function allowing to seize some amount of T
+from the operator whose stake in T contract is lower than their stake in the
+legacy staking contract. 5% of the amount seized is given to the person who
```
vzotova, line 166: Because we can slash only up to allocated amount (per app) then if legacy contract has more stake than allocated it's not malicious behaviour but part of allocating process. Maybe I just misunderstood this sentence

docs/rfc-1-staking-contract.adoc
```diff
@@ -0,0 +1,381 @@
...
+increased.
+
+Effectively, it means that existing KEEP stakers can mix their legacy KEEP
+stakes with liquid T stakes. Similarly, existing NU stakers can mix their legacy
+NU stakes with liquid T stakes. This functionality adds some complexity to the
+staking contract but it puts existing KEEP/NU stakers in the same spot as new T
+stakers. Without it, existing stakers would not be able to top-up their stakes
+with T earned from operating in the network, so they would be in a worse spot
+than new T stakers allowed to top-up their stakes using earned T and this way
+increasing their rewards.
```
eth-r, line 131: If we permit stake-mixing we need to consider what to do when slashing, or when the legacy stake is undelegated but the T stake is not. Permitting the latter case (if the topped-up T stake exceeds the minimum) would let operators move their stakes from legacy to T gracefully without excessive interruption.

docs/rfc-1-staking-contract.adoc
```diff
@@ -0,0 +1,381 @@
...
+undelegation is complete, even before the stake is returned. Information about
+owner, operator, and beneficiary, is retained in the contract in case some
+application rewards are still available for withdrawal.
+
+==== Keeping information in sync
+
+To avoid expensive calls to legacy staking contracts, it is assumed that
+information in T staking contract about the staked amount is always up-to-date.
+By design, this is always the case for native T stakers but might not be the
+case for legacy KEEP/NU stakers.
```
vzotova, line 162: I think no reason to cache whole stake amount, only allocation (for legacy contracts). Because we still don't want to slash more than allocated

docs/rfc-1-staking-contract.adoc
vzotova: Agreed, for PRE app we would need all changes in this amount (probably)

docs/rfc-1-staking-contract.adoc
```diff
@@ -0,0 +1,381 @@
...
+An integral part of the staking contract should be a bot monitoring stakes and
+notifying about discrepancies. This is especially important given that the bot
+may need to voluntarily inform about discrepancies for operators that have been
+slashed to zero.
+
+==== Slashing a stake
+
+Authorized applications can slash or seize a stake. Slash operation decreases
+the stake of an operator and burns slashed tokens. Seize decreases the stake,
+burns 95% of the stake, and awards 5% to the notifier of misbehavior.
```
eth-r, line 201: I'm a bit on the fence about keeping the 5% limit. It's still a good rule to not let any specific recipient get more than 5% of a penalty, but it might be best to enforce this on the application side instead.

If the application can seize the entire penalty amount it can reduce the impact of e.g. TBTC misbehaviour on coverage pools, as seized T could be added to the collateral.

The downside of making the application responsible for the 5% rule is that a malfunctioning application could be used to steal stakes from operators, not just mostly burn them. Being able to limit exposure by authorising less than one's full stake would be able to mitigate this somewhat, but the risk is still there.

eth-r: We also need to consider what to do when slashing legacy/mixed operators. My recommendation would be to slash T first (to avoid legacy contract calls where possible), and convert all seized KEEP/NU to T (so that applications don't need to worry about different tokens).

docs/rfc-1-staking-contract.adoc
```diff
@@ -0,0 +1,381 @@
...
+Allows the governance to approve the particular application before individual
+stake authorizers are able to authorize it.
+
+==== `authorizeApplication(address operator, address application, uint256 amount) external onlyAuthorizerOf(operator)`
+
+Authorizes the particular application for the given operator. From this moment,
+the application may slash operator's stake up to the given amount. Can only be
+called by the given operator's authorizer.
+
+==== `disableApplication(address application) external onlyPanicButton`
```
eth-r, line 258: `onlyPanicButtonOf(application)`?

docs/rfc-1-staking-contract.adoc
pdyraga: I was also thinking about the latter. This could also give tBTC v2/random beacon a chance to update the sortition pool weight. I will clarify.

docs/rfc-1-staking-contract.adoc
```diff
@@ -0,0 +1,381 @@
...
+
+Slashes the provided amount from the stake of every operator in the array.
+Can only be called by an authorized application.
+
+==== `seize(uint256 amount, uint256 rewardMultipier, address notifier, address[] memory operators) external onlyAuthorizedApplication`
+
+Seize the provided amount from the stake of every operator in the array.
+The misbehavior notifier is rewarded with 5% of the total seized amount scaled
+by the reward adjustment parameter. The rest of the tokens are burned. Can only
+be called by an authorized application.
```
eth-r, line 348: We might want to make this seize the tokens for the application instead (and convert legacy to T), regardless of what we do on the 5% rule. In that case it should also return the amount transferred.

docs/rfc-1-staking-contract.adoc
```diff
@@ -0,0 +1,381 @@
...
+
+The difference between eligible and active stake is that the active stake does
+not make the operator eligible for new work selection but it may still finish
+earlier work it is assigned to before the stake is recovered.
+
+==== `activeStake(address operator, address application) external view returns (uint256)`
+
+Returns the active stake amount of the operator for the application. An active
+stake is a stake that may be in the process of undelegation but has not been
+recovered yet. The given application had to be approved for use by the
```
eth-r, line 367: `has not finished undelegating yet`; stake that hasn't been recovered but has passed the undelegation period should not be active

docs/rfc-1-staking-contract.adoc
```diff
@@ -0,0 +1,381 @@
...
+staking contract can record their operator, beneficiary, and authorizer. Since
+these values are read-only, they should be copied from KEEP staking contract to
+T staking contract and NU stakers should set them during the delegation. For
+existing stakes, the entire stake amount is registered in T staking contract.
+
+There are 2 hours to cancel a stake without suffering an undelegation period.
+This functionality is needed e.g. in case of a typo in the operator address.
+Undelegation can not be canceled in the same transaction in which it was
+created. Once the delegation has been canceled, the operator address can be used
+again.
```
eth-r, line 84: We should note that stake that can be cancelled can not be authorised for any application, to prevent risk-and-obligation-free drive-by reward sniping where someone joins a sortition pool for just long enough to grab a reward and then cancels the stake and removes themselves from the pool without suffering an undelegation period.

docs/rfc-1-staking-contract.adoc
pdyraga: Yeah... I was worried it would complicate rewards and have some side effects hard to predict right now. I can imagine someone decreasing their authorization just before group selection and increasing it back afterward. They do not take the responsibility but depending on how we implement continuous rewards, they are rewarded for the higher stake amount for most of the time.

I'd leave it as-is just to do not introduce any new complexity if we agree on it.

docs/rfc-1-staking-contract.adoc
eth-r: It should be `whose stake in the T contract is higher than their stake in the legacy staking contract`; the intent is to cover the situation where someone gets slashed on legacy side but doesn't update their now lower stake to the T contract and thus benefits.

docs/rfc-1-staking-contract.adoc
```diff
@@ -0,0 +1,381 @@
...
+the operator, on behalf of the owner. A beneficiary doesn’t sign or publish any
+protocol-relevant transactions, but any currency or tokens earned by the
+operator will be transferred to the beneficiary.
+
+Authorizer:: An address appointed by an owner to authorize applications on
+behalf of them. An application must be approved by the authorizer before the
+operator is eligible to participate.
+
+Delegated stake:: An owner’s staked tokens, delegated to the operator by the
+owner. Delegation enables KEEP owners to have their wallets offline and their
```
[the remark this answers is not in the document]
eth-r: Yeah, it applies to all owners of staked tokens.

docs/rfc-1-staking-contract.adoc
eth-r: The KEEP staking contract does not support exclusive allocations, so any stake allocations on the T side may get slashed on the legacy KEEP contract. Because of this, the allocations have to be treated as "risk ceilings" where the real value that counts is `min(allocated, staked)`.

For example, Alice could start with 1,000,000 tokens and allocate them all to app X, then get slashed on KEEP and only have 900,000 tokens remaining while the allocation is unchanged at 1,000,000. As a result, we need the T staking contract to correctly calculate Alice's stake as 900,000 after being informed of the slashing.

If we need allocations to be exclusive (if Alice allocates 300,000 tokens to app Y she can only allocate 700,000 tokens to app Z) it needs to be treated as a special case that doesn't support legacy KEEP stakes.

docs/rfc-1-staking-contract.adoc
```diff
@@ -0,0 +1,381 @@
...
+to different operators. For simplicity, the operator address can not be reused
+between delegations even if the delegation has been recovered. The only
+exception is when the delegation has been canceled. To protect against griefing
+by frontrunning stake delegation transactions and reserving operator addresses,
+staking contract should require a governable minimum stake amount for each
+delegation.
+
+Staking contract binds the owner with the operator, beneficiary, and authorizer
+for the given stake delegation and records the stake delegation amount.
+Delegation is set to active after 2 hours, once the cancellation window passes.
```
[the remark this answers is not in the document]
eth-r: If we have a separate undelegation period we should have the 2 hour cancellation window so that mistakenly delegating to the wrong operator address doesn't lock people's capital up for a long time.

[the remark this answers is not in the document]
vzotova: I mean in current PR there is no lock right after stake, lock applies only after second tx with allocation to app

docs/rfc-1-staking-contract.adoc
vzotova: Allocations are not exclusive (#13) and they are already stored, I'm asking about stake amount in KEEP/NU. Even this amount is stored - it can't prevent from described situation and we will have sync/slash mechanism to prevent this.

docs/rfc-1-staking-contract.adoc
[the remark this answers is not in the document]
vzotova: Like two layers: one is deposit/withdrawing tokens (no lock itself). Second layer - allocation/authorization per app. In that case max(any allocation) will be locked. We can consider 2 hours for cancelling allocation but if reason only to save from accident with initial deposit then owner will be ably to withdraw without waiting

docs/rfc-1-staking-contract.adoc
eth-r: With non-exclusive allocations, it is still possible to run into trouble if we don't cache the stake amount locally.

If Alice has allocated the full 1M tokens to two different apps and gets slashed on one, the allocation on the other cannot be reduced to match because the gas cost of doing so would be unbounded with an increasing number of apps (due to having to iterate trough all possible apps and update their allocation if it's higher than the staked amount).

As a result, we need to check both the amount allocated for the app and the amount the operator actually has staked, and use the lower of the two values.

vzotova: Right now we consider 3 apps, iterate through dozens of apps still will be not so heavy. If remove iterations then app can't just wait calls with allocation changes so each time need to check contract again which leads to idea with callback won't work. Also this part is only for legacy stuff so I hope when we will have too much apps - we won't need in this optimizations

docs/rfc-1-staking-contract.adoc
eth-r: This is similar to the stake locking mechanism of the KEEP staking contract. We could let applications set their own deallocation periods, locking the stake so that the application has to release it or be stopped with the panic button before the staked tokens can be withdrawn. In this case it would also be possible to do partial withdrawals of stake by tracking the allocated amounts and requiring that the stake left after withdrawal must be at least as much as the highest allocation.

In this case we wouldn't need a separate undelegation period on the T staking contract.

The prospect of the stake being undelegated on KEEP without being properly deallocated on T can be prevented by the T staking contract locking the KEEP stake until all deallocations on T are completed.

docs/rfc-1-staking-contract.adoc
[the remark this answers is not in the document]
eth-r: Allocations are better controlled (has to be approved by governance etc.) than operator addresses (any arbitrary address works) so I don't think we need a cancellation for allocations.

docs/rfc-1-staking-contract.adoc
eth-r: Iterating through each app is not just an issue with legacy staking, but also any new ones that exclusively support T:

Updating an operator's weight in a sortition pool might cost 50,000 gas as a part of a larger transaction. If we iterate through e.g. 10 allocations and have to perform a callback to update the sortition pools for each of them, it would be 500,000 gas from the pool operations and around 60,000 gas from the allocations on the T staking contract. With 10 operators the cost would be 5.6M gas, and doing it to 100 operators would go way past the gas limit with 56M.

On the other hand, if we only update the stake and let the other applications go out of sync, we can handle 100 operators with around 600,000 gas.

If we need to slash a large number of operators at once, updating the applications becomes prohibitively expensive which is why the new sortition pool design contains a mechanism for incentivising syncing the pool and the staking contract without relying on the staking contract performing callbacks.

docs/rfc-1-staking-contract.adoc
```diff
@@ -0,0 +1,381 @@
...
+Before the operator is eligible to participate in the given application, the
+authorizer appointed during the stake delegation needs to review the application
+and approve it to use the stake. From the moment of approval, the application
+is authorized to slash or seize the stake.
+
+The authorizer decides up to which amount the given application is eligible to
+slash. This allows managing the risk profile of the staker individually for each
+application. The authorized stake amount can only be increased.
+
+Authorization can not be canceled.
```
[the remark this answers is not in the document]
cygnusv: This part needs clarification because it should be possible to deallocate, following pertinent application- and stake-level restrictions (e.g., satisfying specific deallocation periods).

docs/rfc-1-staking-contract.adoc
cygnusv: > The prospect of the stake being undelegated on KEEP without being properly deallocated on T can be prevented by the T staking contract locking the KEEP stake until all deallocations on T are completed.

As far as I understand this is not possible because there's no mechanism for the T staking contract to control stake locks on the KEEP staking contract. On top of that, the latter is not upgradeable. This is essentially the reason why we opened https://github.com/threshold-network/solidity-contracts/issues/14, in agreement with @pdyraga. Can you confirm if this is the case?

docs/rfc-1-staking-contract.adoc
cygnusv: Allowing stake information to go out of sync wrt to applications can break the rewards distribution model that we already agreed on. This is also one of the reasons we opened issue #14, precisely to disallow this possibility.. 

@eth-r Can you clarify on what scenarios there's a need to update 10 or 100 operators' weight at once? Also, I don't understand why sortition pools are being mentioned because they're application-specific (AFAIK, for tBTCv2), and as such, I don't think such restrictions apply to the general T staking model. Similarly, we don't need to consider a model with an unbounded number of apps per stake (we currently only have 3); quite the opposite, we probably should limit it to a sane number that allows to iterate when needed.

cygnusv (commented):
Thanks @pdyraga and @eth-r! Some preliminary comments

docs/rfc-1-staking-contract.adoc
eth-r: The T staking contract needs to be authorised by each KEEP operator who wishes to stake on it or else it can't slash them, and this authorisation would also let the T staking contract set and release stake locks on the KEEP staking contract unless there's something I'm missing.

docs/rfc-1-staking-contract.adoc
eth-r: If we consider a scenario where a tBTC or random beacon group has produced an unauthorised signature, we want to slash the entire group (100 members) because a majority must have been dishonest. Now all these 100 members have their stake reduced. With non-exclusive allocations, this potentially affects the allocation for every other application for every single member.

Sortition pools aren't tBTC-specific, we're also planning on using them for random beacon v2 and quite likely for any future apps with similar needs, because an improved version I'm currently working on solves practically every group selection problem we've had (apart from the one where bulk updating 100 operators costs 5 million instead of 500k gas).

If PRE needs to stay up to date in all circumstances we can afford to special-case it, but I would not tie us down to a model where everything has to stay in sync at all times especially when the number of apps where iteration with callbacks runs OOG can be as low as 2 (e.g. tBTCv2 + random beacon v2: 100 operators get slashed on one app and banned from its sortition pool (>5M gas), have their stake adjusted (>600k gas), which triggers updates on the other application's sortition pool (another >5M gas)).

docs/rfc-1-staking-contract.adoc
pdyraga: I think it makes sense only if this call is executed by an off-chain client. Operator keys should not leave machines clients are deployed on. This is doable but annoying - the client would need to keep an old staking contract binding and check on the start old staking contract and new staking contract. We'd also have to cover the path when the stake was "ported" but undelegated/recovered. How much iffy does it feel?

docs/rfc-1-staking-contract.adoc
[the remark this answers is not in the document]
pdyraga: I'll fix it.

docs/rfc-1-staking-contract.adoc
```diff
@@ -0,0 +1,381 @@
...
+== Specification
+
+=== Functionality
+
+==== Delegating a stake
+
+Tokens are delegated by the owner. During the delegation, the owner needs to
+appoint an operator, beneficiary, and authorizer. The owner may decide to
+delegate just a portion of their tokens. The owner may delegate multiple times
+to different operators. For simplicity, the operator address can not be reused
```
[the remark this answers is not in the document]
pdyraga: The proposition is that operator needs to be unique, the rest (owner, beneficiary, authorizer) can be reused between applications. Here are some examples we observe in practice on Keep mainnet that could help to understand the relationships:
- One owner delegates some part of their stake to one operator and some part of their stake to another operator; authorizer and beneficiary are controlled by the owner and reused in these two delegations.
- One owner delegates some part of their stake to one operator and some part of their stake to another operator; authorizer and beneficiary are controlled by operators so different addresses are used in these two delegations.
- Multiple owners delegate to a staking provider who authorizes on behalf of owners; each owner has their own operator address and beneficiary address but the same authorizer address is used.

> What's the relation between the owner, operator, beneficiary, and authorizer, in a model where an owner has delegated to multiple operators? Isn't it effectively separate stakes?

Correct, these are separate stake delegations. Owner/beneficiary/authorizer can be reused between delegations, operator can not.

> On top of this, how do these roles interact with the model that we agreed on with multiple allocations? Specifically, is operator a stake-level role, or is it more an application-level role? 

It is a stake delegation-level role. One operator can be authorized to operate in multiple applications.

> As far as I understand, if an owner can have multiple operators and allocations, the most natural implications is that each allocation is tied to an operator (and viceversa).

One operator can have multiple allocations. I think the "allocation" term is confusing -  it is not really an allocation but an authorization to use, and slash, up to the given amount of delegated stake. 

If we do allocations, in the strict understanding, applications generating more income would be preferred and other applications - still generating income but not that high compared to others - would be starved.

docs/rfc-1-staking-contract.adoc
pdyraga: Yep! I'll fix it.

docs/rfc-1-staking-contract.adoc
pdyraga: > I'm a bit on the fence about keeping the 5% limit. It's still a good rule to not let any specific recipient get more than 5% of a penalty, but it might be best to enforce this on the application side instead.
> 
> If the application can seize the entire penalty amount it can reduce the impact of e.g. TBTC misbehaviour on coverage pools, as seized T could be added to the collateral.
> 
> The downside of making the application responsible for the 5% rule is that a malfunctioning application could be used to steal stakes from operators, not just mostly burn them. Being able to limit exposure by authorising less than one's full stake would be able to mitigate this somewhat, but the risk is still there.

Same here. My default was to let the application decide and, as you say, instead of burning, donate the coverage pool without minting COV tokens, for example.

The complication though is not only the security model where we block malfunctioning applications from stealing stakes but the construction of Keep staking contract. The option when the application seizes 100% of the amount without burning [is not supported there](https://github.com/keep-network/keep-core/blob/3a4f9ca6ccf4e2713cea52f401a87be49561c592/solidity/contracts/TokenStaking.sol#L413-L425). For this RFC I assumed we want to have an MVP version, good enough to launch and support random beacon, tBTC v2, and PRE, and that we'll introduce additional complexity by upgrades, if necessary.

pdyraga: > We also need to consider what to do when slashing legacy/mixed operators. My recommendation would be to slash T first (to avoid legacy contract calls where possible), and convert all seized KEEP/NU to T (so that applications don't need to worry about different tokens).

Let's discuss it [here](https://github.com/threshold-network/solidity-contracts/pull/15#discussion_r703419028). I am fine with both options. Slashing T first helps avoiding legacy contract calls but slashing KEEP/NU first might easen the transition to T over time.

docs/rfc-1-staking-contract.adoc
pdyraga: Unsupported in Keep staking contract (untada). See [this](https://github.com/threshold-network/solidity-contracts/pull/15#discussion_r704314021) comment please.

docs/rfc-1-staking-contract.adoc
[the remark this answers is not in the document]
pdyraga: I think @vzotova suggests removing the cancelation timeout and let the stake be withdrawn as soon as all authorized applications release it or, if there were no authorizations yet, let to undelegate the stake immediately. If so it depends on the result of [this](https://github.com/threshold-network/solidity-contracts/pull/15/files/c1e8a0ba7479ffb23fd6881d4bc08f113a8a2c2c#r703412443) discussion.

docs/rfc-1-staking-contract.adoc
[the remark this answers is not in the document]
pdyraga: Allowing to cancel the authorization makes the process a little more complex. Canceling authorization per application means that every application should have it is own rules about the "cooldown" period. 

For example, for the random beacon, you need to initialize deauthorization and your stake is no longer eligible for new work selection. After 2 months of deauthorization period, random beacon can no longer slash the stake. 
For tBTC v2, the deauthorization period could be one month and for PRE one week (just an example).
In this model, the stake would get unlocked as soon as the last deauthorization process completes and we wouldn't need a global undelegation period. Deauthorization would happen on the staking contract level and the application would be notified about it. Also, probably we would not need the stake delegation cancellation timeout.

The model presented currently in this RFC is quite similar except that there is a global undelegation time that all applications need to be satisfied with and there is an assumption that authorized application needs to be trusted enough so that when operator stops using it without undelegating, it is not going to slash the stake.

Another - more sophisticated - option would be to let each authorized application lock a stake. Instead of having a hardcoded deauthorization period per application, the application would get notified about deauthorization request and it would trigger an internal cooldown process. Once that process completes - and whatever it means depends on the application - the application would unlock the stake on the staking contract. The stake would get  unlocked as soon as the last lock on the stake is cleared. This option is tempting but it's also way more complex given that each application can receive authorization for a different amount of tokens as [explained here](https://github.com/threshold-network/solidity-contracts/pull/15/files/c1e8a0ba7479ffb23fd6881d4bc08f113a8a2c2c#r703609027) by @eth-r.

docs/rfc-1-staking-contract.adoc
pdyraga: 👌🏻 well said

docs/rfc-1-staking-contract.adoc
pdyraga: Do you mean that the top-up does not automatically increase this possibility given that the higher amount first needs to be authorized? If so, agreed. And I'll clarify it.

docs/rfc-1-staking-contract.adoc
pdyraga: Stake can not be locked in Keep staking contract [for more than 200 days](https://github.com/keep-network/keep-core/blob/3a4f9ca6ccf4e2713cea52f401a87be49561c592/solidity/contracts/libraries/staking/Locks.sol#L19). I think we can implement per-application deauthorization and get rid of stake delegation cancellation timeout and global undelegation time (see [here](https://github.com/threshold-network/solidity-contracts/pull/15/files/c1e8a0ba7479ffb23fd6881d4bc08f113a8a2c2c#r704368646) please) but partial withdrawals are making everything quite complex. I would personally avoid them for the first version of this contract.

docs/rfc-1-staking-contract.adoc
pdyraga: @eth-r is right, this is a typo: s/lower/higher

docs/rfc-1-staking-contract.adoc
eth-r: I guess the fact that T has to be authorised is enough to prevent the minor abuse of people getting randomly staked on T with no control over the process whatsoever, so we actually don't need the access controls here

docs/rfc-1-staking-contract.adoc
eth-r: The stake locks can be renewed but it would probably make things a bit too complex.

Partial withdrawals can also be left for later, agreed.

docs/rfc-1-staking-contract.adoc
eth-r: > For this RFC I assumed we want to have an MVP version, good enough to launch and support random beacon, tBTC v2, and PRE, and that we'll introduce additional complexity by upgrades, if necessary.

I'm fine with this.

docs/rfc-1-staking-contract.adoc
pdyraga: Authorization is just a risk ceiling and the eligible stake for the application is `min(authorization, stake)`. Specifically, the application should not treat the "allocation" (authorization) as a stake; otherwise, we are running into problems described here. 

Caching the entire stake amount feels simpler implementation-wise. For each new authorization (allocation), we wouldn't need to consult the legacy contract and just check the locally cached value assuming it is up to date. If it is not, someone will notify about this fact and the operator will get slashed.

There is an interesting corner case when authorization happens before someone has a chance to notify about discrepancy but I think we can fix it by allowing to notify per application and slashing multiple times. (?)

docs/rfc-1-staking-contract.adoc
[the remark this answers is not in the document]
vzotova: When each app has own cooldown period increases complexity only with iteration over apps to check time and value instead of one value (there is mvp in #7). And we will have panic button so trusting is not so bad at all. From our side we want partial deallocation/deauthorization so anyway it will increase complexity of this process and adding one parameter with time variable is not so big (in my opinion)

pdyraga:
Thanks for the great conversation! I think we are mostly aligned and two major concerns that needs to be addressed first are:
- Is application deauthorization allowed and how should it work; [this](https://github.com/threshold-network/solidity-contracts/pull/15/files/c1e8a0ba7479ffb23fd6881d4bc08f113a8a2c2c#r704368646) thread.
- Is the entire staked amount cached or just the highest authorization; [this](https://github.com/threshold-network/solidity-contracts/pull/15#discussion_r703419120) thread.

These two decisions will have implications on other parts of the spec (cancellation timeout, undelegation period).

docs/rfc-1-staking-contract.adoc
vzotova: Adding one call to authorization is not so expensive considering that authorization calls will be rare comparing other functions. Like for average longtime user it will be few times. And this fully remove issue with not syncing at authorization time. Caching this value and not syncing on slashing is bigger question, I agree

docs/rfc-1-staking-contract.adoc
eth-r: Here would be my proposal for deallocations and keeping applications up to date (including solving [this problem](https://github.com/threshold-network/solidity-contracts/pull/15/files/c1e8a0ba7479ffb23fd6881d4bc08f113a8a2c2c#r703723815)):

Each application has an allocation which is the actual amount staked for it, not just a risk ceiling. Each operator also has a locally recorded stake.

The allocation can be increased by the authoriser, up to the operator's stake. This notifies the application of an allocation increase with `increaseAllocation(address operator, uint256 newAmount)`.

The allocation can be decreased by the authoriser, down to zero. This calls `requestAllocationDecrease(address operator, uint256 newAmount)` on the application and the requested decrease is marked as pending on the staking contract. The application may then call `approveAllocationDecrease(address operator)` at its discretion which sets the allocation to the pending amount.

If the operator's stake falls below the allocated amount (including the currently active allocations even if a decrease request is pending) for any application, anyone can report this to the staking contract which then calls `involuntaryAllocationDecrease(address operator, uint256 newAmount)` on the affected application. The application must handle this immediately. If this is reported within a short grace period of the stake being changed, the operator is not penalised further. If this is reported after the grace period is over, whoever submits the report is rewarded with some stake seized from the operator who failed to report it within time.

Optionally, we could also let operators top up their stake within the grace period and not face a penalty if they bring their stake back to the allocation.

We keep a record of all top-ups and slashings along with the times they happened at, per operator, which can be queried to check the grace period. To prevent spam and OOG attacks on involuntary allocation decrease reporting (not sure _why_ someone would do that but it creates an unbounded iteration) there should be a limit on how many times stakes can be topped up within the grace period, so that a top-up is rejected if e.g. the 20 most recent stake changes have all been after `block.timestamp - gracePeriod`.

When an operator is slashed on T, their stake is updated (along with a record of when the stake changed) but allocations aren't; this ensures that any application that has to slash a large number of operators at once can still predict the gas cost of doing so.

I'm not entirely sure about what to do when an operator is slashed on a legacy contract and it's reported to T; we could either leave it without a reward and use a bot, or have a constant reward (because we can't calculate grace periods) which is large enough that it will probably get collected but creates a race to report.

docs/rfc-1-staking-contract.adoc
eth-r: [From here](https://github.com/threshold-network/solidity-contracts/pull/15#discussion_r704315662)

> Slashing T first helps avoiding legacy contract calls but slashing KEEP/NU first might easen the transition to T over time.

@pdyraga are we expecting honest operators getting slashed to be common enough that we should cater to this specifically?

docs/rfc-1-staking-contract.adoc
eth-r: The proposed mechanism here would essentially move the async discrepancy reporting functionality from the sortition pool to the staking contract, where we can get more information and thus e.g. use penalties with a suitable grace period.

[Question from here](https://github.com/threshold-network/solidity-contracts/pull/15#discussion_r704400260)

> There is an interesting corner case when authorization happens before someone has a chance to notify about discrepancy but I think we can fix it by allowing to notify per application and slashing multiple times. (?)

In my proposal the authorisation would happen, then after the stake has been updated we would have a discrepancy in the operator's authorisation. The grace period would start from when the stake was updated, and the operator can either top up their stake to eliminate the discrepancy, report it themselves, or wait for someone else to report it and grab the reporting reward.

eth-r: One major weakness with this approach is that if an operator gets slashed to zero, there is no way for the staking contract to pay for the reward for reporting a discrepancy, while a sortition pool would be able to use its own maintenance rewards for the purpose. This can be addressed by including the notifier address in the `involuntaryAllocationDecrease()` call, so the application can pay out its own rewards as well.

docs/rfc-1-staking-contract.adoc
pdyraga: Nah, I do not expect any slashes for honest operators. That's actually a good question, I do not think the situation I was worried about is realistic. 

So I think:
- slash T stake first (for simplicity)
- figure out the way to undelegate from the legacy contract without undelegating from the new contract and not being slashed for stake discrepancy 

could be the way to go.

docs/rfc-1-staking-contract.adoc
pdyraga: > Each application has an allocation which is the actual amount staked for it, not just a risk ceiling. Each operator also has a locally recorded stake.

Does it mean that I need to split my stake between several applications? If so, I am worried applications paying smaller rewards can get starved with not enough stakers being available.

Consider the example that random beacon pays 10 T weekly and tBTC v2 pays 100 T weekly for a min stake. Normally, I'd stake for both and earn 110 T but given that I have to choose, I'd go to TBTC v2 with my entire stake .

eth-r: > Does it mean that I need to split my stake between several applications? If so, I am worried applications paying smaller rewards can get starved with not enough stakers being available.

No, the allocations are overlapping so you can allocate the maximum stake to every app

cygnusv: > The allocation can be decreased by the authoriser (possibly also operator?), down to zero. This calls `requestAllocationDecrease(address operator, uint256 newAmount)` on the application and the requested decrease is marked as pending on the staking contract. The application may then call `approveAllocationDecrease(address operator)` at its discretion which sets the allocation to the pending amount. A new allocation decrease cannot be requested while an existing decrease is pending.

@eth-r This part needs further clarification. Can you explain further what you mean by `"The application may then call ...."`, who in particular is performing that call?

docs/rfc-1-staking-contract.adoc
cygnusv: Similarly to another comment I made, the 5% limit seems arbitrary and simply a continuation of KEEP's legacy parameters. We can't foresee if some future application may have stricter seizing requirements. I'd propose to add this as a per-application, governance-level parameter in the T staking contract that's set when the application is connected to the staking contract. If such applications exist at some point, KEEP legacy stakes would not be eligible to participate due to their 5% seize limit, using the mapping we agreed some weeks ago, although it doesn't seem this is currently the case (it's certainly not the case for PRE).

docs/rfc-1-staking-contract.adoc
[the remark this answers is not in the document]
cygnusv: > One owner delegates some part of their stake to one operator and some part of their stake to another operator; authorizer and beneficiary are controlled by operators so different addresses are used in these two delegations.

You lost me here. "authorizer and beneficiary are controlled by operators"? I thought these roles are always controlled by owners. Overall, I still don't get how the multiple operators per stake works in a multi-app setting.

> One operator can have multiple allocations. I think the "allocation" term is confusing - it is not really an allocation but an authorization to use, and slash, up to the given amount of delegated stake.
If we do allocations, in the strict understanding, applications generating more income would be preferred and other applications - still generating income but not that high compared to others - would be starved.

OK, I understand that the term "allocations" could imply exclusive amounts to someone, and this is not the case. Let's go with "authorizations" 👍

docs/rfc-1-staking-contract.adoc
eth-r: The staking contract would not care about what happens on the application side, all it requires is that `approveAllocationDecrease()` is called by the application.

In practice we could e.g. have the application start a timer when it receives the `requestAllocationDecrease()` call and when the required time has passed the operator can notify the application of this which then makes the application call the staking contract.

so the request would go `operator -> staking -> application` and the approval would go `operator -> application -> staking` where `a -> b` means that `a` calls a function on `b`

docs/rfc-1-staking-contract.adoc
eth-r: We could make it so that T will be seized 100% and legacy stake is limited to the 5%. Thus applications that need the full seizure functionality would only use T stakes (with the other benefits, such as staying better synced) and how they distribute the seized tokens would be an internal matter.

docs/rfc-1-staking-contract.adoc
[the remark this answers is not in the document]
eth-r: > You lost me here. "authorizer and beneficiary are controlled by operators"? I thought these roles are always controlled by owners. Overall, I still don't get how the multiple operators per stake works in a multi-app setting.

There would not be multiple operators per stake. The same _owner_ can stake tokens at multiple operators, where each operator is an independent stake for all purposes. Stakes are uniquely associated with the operator addresses, and the staking contract doesn't care about the owner address except when returning the staked tokens.

The authoriser and beneficiary are roles that can be freely set by the owner when staking tokens. The authoriser and beneficiary can be controlled by either the owner or the operator, the staking contract doesn't care about the off-chain arrangements.

[Here is the original explanation of the roles](https://github.com/keep-network/keep-core/blob/main/docs/random-beacon/staking/index.adoc#roles)

docs/rfc-1-staking-contract.adoc
eth-r: If we permit authorisation decreases, we could require that the operator has reduced their authorisations on all apps to or below the T stake they have before being allowed to initiate undelegation on KEEP. Undelegating without reducing the authorisations first would be treated as a discrepancy.

pdyraga: Good idea!

docs/rfc-1-staking-contract.adoc
cygnusv: I think this could be achieved without that second call. When staker calls `requestAllocationDecrease(address operator, uint256 newAmount)`, the staking contract will store the new amount and a future timestamp with the time when tokens will be effectively out of the authorization; additionally it performs a callback to the application in case it should reject the allocation decrease or some other specific logic. Later, if the staker wants to withdraw or make any other change to their stake, we check the pending decreases in the staking contract.

docs/rfc-1-staking-contract.adoc
cygnusv: Yes, that's roughly what we agreed o with @pdyraga during last call. There's a mapping in the staking contract that specifies which apps are compatible with the possible "stake provider" (i.e., native T, KEEP legacy, NU legacy), so we'd use this to allow/restrict KEEP legacy stakes as needed.

docs/rfc-1-staking-contract.adoc
pdyraga: Let me try to rephrase it and compare it to the original proposition to make sure we are all on the same page.

The original proposition was: the staked amount is a staking-contract-level variable, `staked`. Each application receives an individual authorization, which is a risk ceiling for the given application, `authorized_app1`, `authorized_app2`, and so forth. In this model, `eligibleStake(appN) = min(staked, authorized_appN)`. It means that we do not need to update the authorized amount for each application if one application slashed the `stake`. When someone authorizes an application, we do validate if `authorized <= stake`, but we do not guarantee this will always be the case in the future. Thus, we have to calculate `eligibleStake` as a function of `staked` and `authorized_appN`.

To keep it simple, the original proposition assumed lowering `authorized_appN` is not possible and there is no way to lower the risk ceiling for applications without undelegating and that the undelegation period is global, hardcoded in the staking contract, and all applications need to be designed in a way obeying the undelegation period. 

We decided that it would be nice - or maybe even a must-have - to have a possibility to decrease the authorization (allocation) from a particular application. Quite a logical implication would be to allow each application to have its own cooldown rules for deauthorization (uncallocation).

I *think* we are not that far in the original proposition from having this feature and we do not have to store the current allocation per application and introduce additional mechanisms to keep them up to date. Trying to modify @eth-r's proposal: 
- `staked` amount is still a staking-contract-level variable
- each application receives an individual authorization (risk ceiling), `authorized_appN`
- `eligibleStake(appN) = min(staked, authorized_appN)`, 
- we probably need a separate `authorized_appN` flavor to reflect the pending authorization decrease for the given app to be able to calculate `activeStake`
- authorizer calls `requestAuthorizationDecrease` on the staking contract, staking contract executes a call to the application so that it can update internal state or even revert the transaction if the given application does not support decreasing the authorization
- after some time, the application calls `finalizeAuthorizationDecrease` that updates `authorized_appN` value on the staking contract; the time before this call is performed by the application, and everything that needs to happen before the process is finalized depends on the particular application.

docs/rfc-1-staking-contract.adoc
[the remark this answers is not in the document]
pdyraga: (changing this mechanism is discussed in [this](https://github.com/threshold-network/solidity-contracts/pull/15#discussion_r703412443) thread)

docs/rfc-1-staking-contract.adoc
pdyraga: In the Limitations section, there is a note we do not guarantee legacy stakes will be supported in all future applications. 

I proposed to keep the limit of 5% for the first applications so that existing Keep stakers are in a good spot after the network launch and their stake is honored by all applications launching in T for some time; implicitly it means applications need to accept 5% is the limit. Then, once everyone has a chance to earn some T to migrate, we will update the staking contract and remove 5% limit. 

I think this approach has better optics for stakers and is simpler to implement for the first version of staking contract.

If we want to remove the limit now, I think we have two options:
- store a mapping of applications supporting legacy stake in the contract (as currently proposed in the draft implementation)
- provide enough information during `authorizeApplication` so that the application can decide (is it legacy stake, if so in what proportions given that we allow mixing legacy and T stake)

I think in terms of the complexity: accept 5% limit, then upgrade is easier than mapping and mapping is easier than a smart `authorizeApplication` call.

docs/rfc-1-staking-contract.adoc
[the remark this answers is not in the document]
pdyraga: Just like @eth-r explained. I think one more thing is worth adding here as I suspect this could be confusing: operator != staking provider. An operator is just an address, same beneficiary, authorizer, and owner. From the perspective of the staking contract, who controls them does not matter. Practically, we've seen different setups: individuals controlling all 4 roles, owners delegating to staking providers but still controlling beneficiary and authorizer, owners delegating to staking providers and using beneficiary and authorizer controlled by them.

docs/rfc-1-staking-contract.adoc
cygnusv: > Let me try to rephrase it and compare it to the original proposition to make sure we are all on the same page.
> 
> The original proposition was: the staked amount is a staking-contract-level variable, `staked`. 

Right, although that's under a model where we allow authorized amounts to go out of sync. We're not necessarily pushing for a different model right now, but it's important to note that if we keep them on sync, we don't need this variable.

> Each application receives an individual authorization, which is a risk ceiling for the given application, `authorized_app1`, `authorized_app2`, and so forth. 

Right, although the fact that's a risk ceiling is from the stake contract point of view.  From the applications point of view, the authorized amount is effectively the "stake" (let's call it "application stake" to not confuse it with the principal stake) and it's used not only as a measure of risk (e.g. for slashing purposes), but also as a basis for rewards, fees, or other potential application logic. 

> In this model, `eligibleStake(appN) = min(staked, authorized_appN)`. It means that we do not need to update the authorized amount for each application if one application slashed the `stake`. When someone authorizes an application, we do validate if `authorized <= stake`, but we do not guarantee this will always be the case in the future. Thus, we have to calculate `eligibleStake` as a function of `staked` and `authorized_appN`.

But if we are operating under the assumption that we allow some synchronization flexibility, there's no need for the application to read the `staked` variable to compute the eligibleStake. Also, we still need to answer the big question of how are we going to ensure some degree of synchronization between app and staking contracts after some time.

> To keep it simple, the original proposition assumed lowering `authorized_appN` is not possible and there is no way to lower the risk ceiling for applications without undelegating and that the undelegation period is global, hardcoded in the staking contract, and all applications need to be designed in a way obeying the undelegation period.
> 
> We decided that it would be nice - or maybe even a must-have - to have a possibility to decrease the authorization (allocation) from a particular application. Quite a logical implication would be to allow each application to have its own cooldown rules for deauthorization (uncallocation).

Yes, the fact that authorizations are non-exclusive directly implies a higher profits and risks margin for the same underlying stake, so it makes sense to allow some risk reduction mechanism.

docs/rfc-1-staking-contract.adoc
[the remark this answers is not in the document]
pdyraga: https://github.com/threshold-network/solidity-contracts/pull/15/commits/ff2ff637b758853aa667b34a2dfccbd9af048a41

pdyraga:
I updated the RFC to reflect everything we have discussed recently and things that I think makes sense given the new ideas:
- Removed stake delegation cancel functionality in favor of just undelegate with a minimum stake and 24 minimum undelegation delay to protect against griefing by front-running delegation transactions and reserving operator addresses. Undelegation delay is no longer needed - each application needs to have their authorization decreased to 0 first.
- Added a note on caching legacy stake amount in T staking contract.  It is much easier to just compare two values instead of checking authorization of every application. Some Ethereum providers charge per compute unit so the simpler the check, the more convenient it is for all network operators to monitor and report the discrepancy.
- Authorization (still using this word because IMO allocation suggests the amount can not be reused between applications) can now be increased and decreased. Authorization increase is a one-step process and authorization decrease is a two-step process finalized by the application at its discretion.
- Undelegation can happen immediately if all authorizations have been zeroed. 
- Introduced slashing queues with a note of 250k gas limit for each application callback and not reverting the transaction if an individual callback failed.
- There is now an option to unstake KEEP and unstake NU while still operating in T and earning rewards. One who accumulated enough T to run a node can have a mixed KEEP<>T / NU<>T stake and lower the cached legacy stake amount to 0 in T staking contract.

I think all comments have been addressed.

docs/rfc-1-staking-contract.adoc
```diff
@@ -132,47 +145,44 @@ increasing their rewards.
...
-
-Once the undelegation period finishes, anyone - even a third party - can recover
-the stake. Staking contract sets the stake as recovered and returns all
-delegated tokens to the owner. Applications shouldn't be able to slash after
-undelegation is complete, even before the stake is returned. Information about
-owner, operator, and beneficiary, is retained in the contract in case some
-application rewards are still available for withdrawal.
+The owner or operator may choose to undelegate the stake if there are no
+authorizations on applications or if all authorizations have been decreased to
+zero. It is not possible to partially undelegate the stake. Undelegation returns
```
vzotova, line 154: Can we allow partial undelegation? It would be useful, for example, when staker has mixed stake and want to withdraw only NU and can withdraw only partial amount. First deauthorize this amount of course.

docs/rfc-1-staking-contract.adoc
```diff
@@ -185,20 +195,56 @@ is higher than the stake amount on T staking contract.
...
-slashed to zero.
+An integral part of the staking contract should be a bot or process inside an
+off-chain client monitoring stakes and notifying about discrepancies. This is
+especially important given that the bot may need to voluntarily inform about
+discrepancies for operators that have been slashed to zero.
+
+Owner or operator can decrease the legacy contract active stake cached amount on
+T staking contract if no application has authorization higher than the liquid T
+stake. It allows to undelegate from the legacy staking contract while still
+being able to operate in T and earn rewards.
```
vzotova, line 210: What do you think about giving control on T, cached NU/KEEP separately in T staking contract? We can have some default rules when authorize T first, deauthorize NU/KEEP first and give ability to set this when needed?

docs/rfc-1-staking-contract.adoc
```diff
@@ -185,20 +195,56 @@ is higher than the stake amount on T staking contract.
...
+affected application, and notifying all affected applications that
+the operator's authorized stake has been reduced due to slashing. The
+application must then do the necessary adjustments, such as removing the
+operator from the sortition pool or reducing its weight, changing the operator's
+eligibility for rewards, and so forth.
+
+Every application callback executed as a result of slash should have a 250k gas
+limit. Slashing are processed in a FIFO basis, and there is just one function
+exposed by the staking contract allowing to slash one or more operators from the
+head of the queue. Callback failure does not revert the transaction.
```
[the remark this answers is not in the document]
pdyraga: I think one possibility would be to move the problematic slashing operation to the end of the queue and take another one from the front of the queue in the same transaction. We'd need to ensure that the operation moving elements in the queue is really cheap so that the cost of it does not exceed gains from executing the slashing.

This has a disadvantage though - in case the application is misbehaving and no slashes from the queue can be executed they keep growing and clogging the queue. One misbehaving application can render slashing of all other applications impossible to execute. Now, if the application can not properly update its internal state, should we let this problem propagate to other applications? This problem could be partially resolved by having the panic button clearing up the queue from slashings of misbehaving application, but before the panic button can disable application slashing and clear the queue, a lot of bad things can happen. This approach also has the disadvantage that the panic button triggered in other circumstances would let some operators benefit from the timing.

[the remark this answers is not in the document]
eth-r: I think misbehaving applications shouldn't be able to hold up the slashing, and each operator should be cleared from the queue by giving each application one chance to update. If we don't want to just gloss over errors in updating applications we could have another "problem list" where we move the `(operator, application)` pair of each slashing notification call that fails. The problem list would permit notifications on a pairwise basis (so you can call `problemNotify(operator, applicationA)` and it updates application A but leaves application B not updated) for cleaning up, with an arbitrary number of repeat attempts with the pair only being removed from the list once the call is successful. No need for any kind of rewards or other incentives here, just something to facilitate housekeeping.

In the case of a panic buttoned application, the slashing processing should just skip it without calling or adding the operator+application pair to the problem list. Similarly, if a pair on the problem list contains an application that has been panic buttoned, the pair can be simply removed from the list without making a call to the panic-buttoned application.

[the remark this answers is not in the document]
pdyraga: A separate queue for problematic slashings was also proposed by @vzotova today during the weekly call. I can see this idea working but the more I think about it the less I am convinced it is worth the additional complexity. It is in the best interest of the application requesting the slash to be able to process the slashing request from the queue. If the application can not handle it, the application is only hurting itself. I'd vote for giving just one chance to update the state for each application (with a gas cap) and if the application can not process it, the requested update is ignored.

docs/rfc-1-staking-contract.adoc
pdyraga: What I think we can do is to eliminate `undelegate(address operator) external` and replace it with:
- `unstakeKeep(address operator)` - sets cached legacy KEEP stake amount to 0; reverts if there is at least one authorization higher than the liquid T stake for that operator or if there was no KEEP staked for the operator. If succeeds, the legacy stake can be undelegated.
- `unstakeNu(address operator, uint256 amount)` - reduces cached legacy NU stake amount by `amount`; reverts if there is at least one authorization higher than the sum of remaining legacy NU stake and liquid T stake for that operator or if `amount` is higher than the cached legacy stake amount. If succeeds, the legacy NU stake can be partially undelegated.
- `unstakeT(address operator, uint256 amount)` - reduced the liquid T stake amount by `amount` and withdraws `amount` of staked T to the owner; reverts if there is at least one authorization higher than the sum of a legacy stake and remaining liquid T stake or if `amount` is higher that the liquid T stake amount.
- `unstakeAll(address operator)` - sets cached legacy stake amount to 0, sets the liquid T stake to 0 and withdraws all staked T to the owner; reverts if there is at least one non-zero authorization.

What do you think?

docs/rfc-1-staking-contract.adoc
pdyraga: I do not see any obvious advantage of allowing it. With the existing top-up API and partial undelegations proposed in the [comment above](https://github.com/threshold-network/solidity-contracts/pull/15#discussion_r711199239), stakers have full control over their portfolio: what is the share of legacy stake and what is the share of T stake. They can partially withdraw or even eliminate the legacy stake from their portfolio entirely while still operating a node and earning rewards.

Given that - for simplicity - we always want to slash T stake first, allowing to split authorizations between legacy stake and T stake, would make the slashing quite complex.

docs/rfc-1-staking-contract.adoc
vzotova: Agree with idea, but few comments about naming:  in case legacy stakes `unstake` means save it from slashing but withdrawing will be outside t staking contract, when `unstake` for T includes withdraw. So maybe have different names and do not provide united method (just suggestion)?

docs/rfc-1-staking-contract.adoc
```diff
@@ -0,0 +1,479 @@
...
+Allows the governance to set the minimum required stake amount. This amount is
+required to protect against griefing the staking contract and individual
+applications are allowed to require higher minimum stakes if necessary.  
+
+=== Authorizing an application
+
+==== `approveApplication(address application) external onlyGovernance`
+
+Allows the governance to approve the particular application before individual
+stake authorizers are able to authorize it.
```
vzotova, line 298: Do we want to have mapping or list in input parameters to specify which stakes are approved (T, KEEP, NU or all of them)?

docs/rfc-1-staking-contract.adoc
vzotova: Question regarding this method (also related to `stakeNu`):
what if operator is already in use (by staking T previously) and/or beneficiary and authorizer are different comparing between Keep contract values and T contract values? Do we want to reject tx and if so what to do with possibility stealing operator address?

docs/rfc-1-staking-contract.adoc
vzotova: I'm ok with `unstake` and putting difference in docs. Also maybe some complex name like `unstakeAndWithdraw` for T?

pdyraga: I was thinking about `unstakeAndWithdraw` but the complication is with `unstakeAll` that is withdrawing just T and leaving the rest of withdrawal to legacy contracts.

docs/rfc-1-staking-contract.adoc
vzotova: I am definitely for simplicity and not worrying about availability on start. 
Just think that all logic could be included in staking contract and not so big. Because when app is not available for some legacy stake that means when authorization increased - ceil of authorization is lower (same for slashing/syncing). App would work only with authorization knowing nothing about the type of stake. 
But again I totally agree not to implement this on day one

docs/rfc-1-staking-contract.adoc
vzotova: `unstakeAllAndWithdrawT` ? 😄

docs/rfc-1-staking-contract.adoc
pdyraga: To explain the problematic case I have in mind:
- Operator has 100k KEEP and 1M T staked. For simplicity, let's assume 1 KEEP = 1 T. In total, it's 1.1M `amount` and applications has no idea what is the composition of tokens (applications shouldn't care about it). Now, should we let this operator to authorize application X to `amount = 100k` if it refuses to work with KEEP staking contract? Yes, because the operator has 1M T. So the authorization for that particular application could be for 1M max. 
- Now, that operator calls `unstakeT(amount = 1M)`. We will notify the application `amount` changed, and that it's 100k now but that application is not aware that amount is only in KEEP.

What should happen in step 2 is that the application should revert the TX. But it means applications need to be aware of the composition of stakes. Alternatively, the staking contract would need to check applications, see what stakes they are authorized for, and update only in specific circumstances.

This is getting quite complex in my opinion and T staking contract is already pretty complex. I am afraid, we could be designing for a situation that will never happen given that applications for the next few years could be perfectly fine with 5% notifier reward limit. If we agree on respecting this limit, we are eliminating some complexity from the implementation and sending a friendly signal to existing Keep stakers who cannot unstake from tBTC v1 right now.

vzotova: I see that in `unstakeT` won't be calls to app in anyway because authorization remains the same.
Hypothetical how it could be implemented in staking contract v2: When we will calculate how many tokens reallly can be unstaked - we can see that amount for this app was 100k from 1M possible which means only 900k can be unstaked safely. So app knows nothing, staking contract reverts tx

pdyraga: > I see that in unstakeT won't be calls to app in anyway because authorization remains the same.

Yes, not in the current version of the RFC. I meant that we could add it - just one of the options.

> Hypotheticaly how it could be implemented in staking contract v2: When we will calculate how many tokens reallly can be unstaked - we can see that amount for this app was 100k from 1M possible which means only 900k can be unstaked safely. So app knows nothing, staking contract reverts tx
 
Right, but in this case, the staking contract needs to know what type of stake the given application is willing to accept and validate the current composition each time `unstake*` happens. It is definitely possible to implement, I am just concerned if we really need it given the contract is upgradeable.

vzotova: > Yes, not in the current version of the RFC. I meant that we could add it - just one of the options.

Could you elaborate? In current version decreasing authorization and unstaking are not possible in one tx (because of backward call from app). Or you mean something else?

> Right, but in this case, the staking contract needs to know what type of stake the given application is willing to accept and validate the current composition each time `unstake*` happens. It is definitely possible to implement, I am just concerned if we really need it given the contract is upgradeable.

👍

pdyraga: > Could you elaborate? In current version decreasing authorization and unstaking are not possible in one tx (because of backward call from app). Or you mean something else?

I meant that if we want to have the application aware of the stake type, we could add one more callback to `unstake` to give the application a chance to take a look at the current stake composition and revert the transactions. I hate this idea but this was one of the options I had in mind so posted it here and... caused confusion. 

It seems we agree on starting simple, so let's just forget about this crazy idea of an application callback from `unstake`.

pdyraga: Anyway, feels that we are aligned on everything so I'll work on updating the RFC to a final-ish form tomorrow!

docs/rfc-1-staking-contract.adoc
```diff
@@ -0,0 +1,479 @@
...
+Authorizes the particular application for the given operator. From this moment,
+the application may slash operator's stake up to the given amount. Can only be
+called by the given operator's authorizer.
+
+==== `increaseAuthorization(address operator, address application, uint256 amount) external onlyAuthorizerOf(operator)`
+
+Increases the authorization of the given operator for the given application by
+the given amount. Calls `authorizationIncreased(address operator, uint256 amount)`
+callback on the given application to notify the application. Can only be called
+by the given operator's authorizer.
```
vzotova, line 311: What do you think about to have only one method instead of `increaseAuthorization` and `authorizeApplication`? In current version of doc we have only one callback for increasing authorization. If there will be separate method for initial authorization then maybe add another callback for it?

docs/rfc-1-staking-contract.adoc
vzotova: Another question: looks like different apps can have different minimum authorization (>= min stake). maybe add to `approve` logic also caching this parameter? Or maybe other approach?

docs/rfc-1-staking-contract.adoc
pdyraga: Good call. I think we should drop `authorizeApplication` . `increaseAuthorization` is already covered with the callback and increasing from 0 is still increasing.

pdyraga:
I updated the RFC to the latest state to reflect all the discussions we have had. I'd love if you could do a final pass and if you approve, merge this PR. Some small tweaks can be done in separate PRs where we'd also have a chance to discuss the changes.

docs/rfc-1-staking-contract.adoc
pdyraga: Not sure if I follow. According to the RFC, this is currently handled in the callback that is called for increase and decrease operations and that callback can revert the transaction if some conditions are not met. How would the change you described look like?

vzotova: Nothing special, just cache min amount per app in staking contract and check this beforehand during authorization. Not necessary of course

eth-r (approved):
Looks good.

PR #17, opened 2021-09-23
IApplication and IStaking interfaces

pdyraga:
Refs #1, #15

Solidity interfaces for application and staking contract based on the
functions defined in RFC 1. Having the interfaces lets to unblock work
on applications (random beacon v2, tBTC v2) while the actual
implementation of the staking contract can happen in the meantime.
For the implementation, please see https://github.com/threshold-network/solidity-contracts/pull/7.

Note that nothing is carved in stone and some changes are still
possible. The RFC 1 was thoroughly discussed but some new findings during
the contract implementation time are possible.

### DIFF
--- a/contracts/staking/IStaking.sol
+++ b/contracts/staking/IStaking.sol
@@ -1,0 +1,252 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+pragma solidity 0.8.4;
+
+/// @title Interface of Threshold Network staking contract
+/// @notice The staking contract enables T owners to have their wallets offline
+///         and their stake operated by operators on their behalf. All off-chain
+///         client software should be able to run without exposing operator’s
+///         private key and should not require any owner’s keys at all.
+///         The stake delegation optimizes the network throughput without
+///         compromising the security of the owners’ stake.
+interface IStaking {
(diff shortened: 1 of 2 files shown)


contracts/staking/IStaking.sol
```diff
@@ -0,0 +1,252 @@
...
+
+    /// @notice Adds operators to the slashing queue along with the amount that
+    ///         should be slashed from each one of them. Can only be called by
+    ///         application authorized for all operators in the array.
+    function slash(uint256 amount, address[] memory operators) external;
+
+    /// @notice Adds operators to the slashing queue along with the amount,
+    ///         reward multiplier and notifier address. The notifier will
+    ///         receive 1% of the slashed amount scaled by the reward adjustment
+    ///         parameter once the seize order will be processed. Can only be
```
eth-r, line 216: We need to check the stake of each operator to ensure that we can do this safely; it is possible that a naive implementation would result in an inconsistent state where the staking contract doesn't maintain full reserves.

vzotova (approved):

pdyraga:
Are we good to merge this one?

cygnusv (approved):

PR #21, opened 2021-10-21
Basic governance contracts for Tokenholder DAO and Staker DAO

cygnusv:
Threshold DAO
This PR creates the basic contracts that implement the bicameral governance system of the Threshold Network:
* The Tokenholder DAO controls the main aspects of governance, like T token emissions and governance settings. It uses both liquid and staking voting power sources, namely, the T token and the TokenStaking contracts.
* The Staker DAO controls the staking and application contracts. It uses only a staked voting power source.
* Upgradeability aspects are left for a future PR.

Further design restrictions:
* There's a Council Multisig that can veto proposals in both DAOs
* The Tokenholder DAO controls governance settings of the Staker DAO

Implementation aspects:
* Uses OpenZeppelin's Governance framework as a starting point, reusing when possible, but there's a fair amount of new code.
* Put Timelocks in place to add a risk mitigation layer to the governance process.

Comments for future PRs:
* We will use the Transparent Upgradeable Proxy approach for upgradeability of the TokenStaking contract and application contracts.
* Normally, TokenStaking upgrades will be performed by the Staker DAO, but there's an Emergency Security Developer Multisig (ESDM) that can push urgent upgrades to the staking contract if necessary.
* The Transparent Upgradeable Proxy approach requires the existence of a ProxyAdmin contract that relays upgrade requests, and given that we need an additional emergency upgrade track for the ESDM, we will implement a new contract based on OpenZeppelin's ProxyAdmin called ProxyAdminWithDeputy, with support for this emergency track. In this case, the ESDM will act as the "deputy".
* TODO: Documentation & Deployment scripts

### DIFF
--- a/contracts/governance/StakerGovernor.sol
+++ b/contracts/governance/StakerGovernor.sol
@@ -3,6 +3,7 @@
 pragma solidity 0.8.9;
 
 import "./StakerGovernorVotes.sol";
+import "./TokenholderGovernor.sol";
 import "../token/T.sol";
 import "@openzeppelin/contracts/access/AccessControl.sol";
 import "@openzeppelin/contracts/governance/Governor.sol";
@@ -27,9 +28,12 @@ contract StakerGovernor is
     bytes32 public constant VETO_POWER =
         keccak256("Power to veto proposals in Threshold's Staker DAO");
 
+    address internal immutable manager;
+
     constructor(
         IVotesHistory _staking,
         TimelockController _timelock,
+        TokenholderGovernor tokenholderGovernor,
         address vetoer
     )
         Governor("StakerGovernor")
@@ -42,8 +46,15 @@ contract StakerGovernor is
         StakerGovernorVotes(_staking)
         GovernorTimelockControl(_timelock)
     {
+        require(
+            keccak256(bytes(tokenholderGovernor.name())) ==
+                keccak256(bytes("TokenholderGovernor")),
+            "Incorrect TokenholderGovernor"
+        );
+        manager = tokenholderGovernor.timelock();
+        require(manager != address(0), "No timelock founds");
         _setupRole(VETO_POWER, vetoer);
-        _setupRole(DEFAULT_ADMIN_ROLE, address(_timelock));
+        _setupRole(DEFAULT_ADMIN_ROLE, manager);
     }
 
     function propose(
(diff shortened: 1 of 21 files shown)


contracts/governance/TokenholderGovernor.sol
```diff
@@ -0,0 +1,143 @@
...
+import "@openzeppelin/contracts/governance/extensions/GovernorVotes.sol";
+import "@openzeppelin/contracts/governance/extensions/GovernorTimelockControl.sol";
+
+contract TokenholderGovernor is
+    AccessControl,
+    Governor,
+    GovernorCountingSimple,
+    TokenholderGovernorVotes,
+    GovernorTimelockControl
+{
```
vzotova, line 18: Maybe possible to extract some logic from this and `StakerGovernor` to some other abstract contract?

vzotova (commented):
Looks good to me for now, need more time to understand hierarchy of contracts. Here some questions

contracts/governance/StakerGovernor.sol
```diff
@@ -0,0 +1,143 @@
...
+import "@openzeppelin/contracts/governance/extensions/GovernorVotes.sol";
+import "@openzeppelin/contracts/governance/extensions/GovernorTimelockControl.sol";
+
+contract StakerGovernor is
+    AccessControl,
+    Governor,
+    GovernorCountingSimple,
+    StakerGovernorVotes,
+    GovernorTimelockControl
+{
```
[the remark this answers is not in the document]
cygnusv: Oh thanks, it's actually redundant. I removed it.

yarn.lock
```diff
@@ -551,3501 +551,3647 @@
...
-    js-yaml "^3.14.0"
+    "axios" "^0.21.1"
+    "fs-extra" "^9.0.1"
+    "js-yaml" "^3.14.0"
 
 "@thesis/solidity-contracts@github:thesis/solidity-contracts#4985bcf":
-  version "0.0.1"
-  resolved "https://codeload.github.com/thesis/solidity-contracts/tar.gz/4985bcfc28e36eed9838993b16710e1b500f9e85"
+  "resolved" "git+ssh://git@github.com/thesis/solidity-contracts.git#4985bcfc28e36eed9838993b16710e1b500f9e85"
+  "version" "0.0.1"
```
michalinacienciala: It looks that this change causes the failures of GH Actions checks. Was the `yarn.lock` correctly generated?

cygnusv: Oh thanks for looking into that. I'll try to regenerate the lock file.

vzotova (approved):

PR #22, opened 2021-10-27
Contracts deployment script

michalinacienciala:
This commit introduces a deployment script based on the
[hardhat-deploy](https://github.com/wighawag/hardhat-deploy) plugin.
To deploy all contracts on the given network, the following command can
be used:
```
yarn deploy --network <network>
```
If contracts haven't been built yet or changes occurred, this task will
build the contracts before running the deployment script. Once the
deployment terminates, a new `deployments` directory containing all
deployment info will be created. It can be directly used by dApps or
other client code as it contains deployment details like chain ID,
transaction hash, ABI or address for each contract. There is also a
possibility to run the deployment with export mode enabled:
```
yarn deploy --network <network> --export ./export.json
```
This command produces a handy summary file containing deployment info
for all contracts in one place. However, it doesn't contain the
deployment transaction hash making it inappropriate for some use cases
relying on this field.

The deployment script is divided into multiple sub-scripts placed in the
`deploy` directory. It uses the [tags and dependencies](https://github.com/wighawag/hardhat-deploy#deploy-scripts-tags-and-dependencies)
feature provided by the `hardhat-deploy` plugin. Such a structure
allows running arbitrary parts of the entire deployment by using the tag
mechanism. For example, to deploy only the `VendingMachine` contract (with
their dependencies), the following command can be used:
```
yarn deploy --network localhost --tags VendingMachine
```
Multiple deployment sub-scripts also improves the readability and allows
specifying dependencies between components in an explicit way.

PR #23, opened 2021-10-28
Configure jobs deploying & publishing contracts

michalinacienciala:
We're adding jobs which will deploy and publish Threshold contracts.
`NPM` workflow will build & publish contracts every time there is a
change in contracts or in the files describing package dependencies.
The `-dev` suffix will be used in the name of the published package.
`Solidity` workflow will deploy and publish contracts every time
worklflow is triggered by the `workflow_dispatch` event. Published
package name will point to name of the environment on which the
contracts were deployed.

Note: we're temporarily switched off publishing to NPM registry using
`--dry-run` flag. The flag will be removed before merging of the changes
to the `main` branch.

TODO:
- [x] setup GH secrets
- [x] test the workflows
- [x] cleanup of test configuration

Ref:
https://github.com/threshold-network/solidity-contracts/pull/22

### DIFF
--- a/.github/workflows/contracts.yaml
+++ b/.github/workflows/contracts.yaml
@@ -20,6 +20,23 @@ on:
         default: "main"
 
 jobs:
+  contracts-detect-changes:
+    runs-on: ubuntu-latest
+    outputs:
+      system-tests: ${{ steps.filter.outputs.system-tests }}
+    steps:
+      - uses: actions/checkout@v2
+        if: github.event_name == 'pull_request'
+
+      - uses: dorny/paths-filter@v2
+        if: github.event_name == 'pull_request'
+        id: filter
+        with:
+          filters: |
+            system-tests:
+              - './contracts/staking/**'
+              - './test/system/**'
+
   contracts-build-and-test:
     runs-on: ubuntu-latest
     steps:
@@ -27,7 +44,7 @@ jobs:
 
       - uses: actions/setup-node@v2
         with:
-          node-version: "14"
+          node-version: "14.x"
           cache: "yarn"
 
       - name: Install dependencies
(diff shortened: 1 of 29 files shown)


.github/workflows/contracts.yaml
```diff
@@ -27,8 +39,135 @@ jobs:
...
+      - name: Verify contracts on Etherscan
+        env:
+          ETHERSCAN_API_KEY: ${{ secrets.ETHERSCAN_API_KEY }}
+          CHAIN_API_URL: ${{ secrets.ROPSTEN_ETH_HOSTNAME_HTTP }}
+        # run: |
+        #   yarn run hardhat --network ${{ github.event.inputs.environment }} \
+        #     etherscan-verify --license MIT
+        run: |
+          yarn run hardhat --network ropsten \
+            etherscan-verify --license MIT
```
michalinacienciala, line 166: @Kuba, `T` and `VendingMachine` contracts use `GPL-3.0-or-later` licens, which is not supported by etherscan. Do you think I could force `MIT license here (`yarn run hardhat --network ropsten etherscan-verify --license MIT --force-license`)? I verified that it works locally. Without the `--force-license` or `--license MIT --force-license` I was getting errors preventing from verification).

nkuba: Looks like GPL v3 should be supported by etherscan: https://etherscan.io/contract-license-types.
I'm not sure what value the command is expecting though.

michalinacienciala: I have force using `GPL-3.0` license (`yarn run hardhat --network ropsten etherscan-verify --license GPL-3.0 --force-license`), see cfaa9e1.

michalinacienciala:
Workflows positively tested with testing configuration, see here: https://github.com/threshold-network/solidity-contracts/runs/4210925769?check_suite_focus=true.
I removed the testing configuration, PR is ready for review.

nkuba (approved):

PR #25, opened 2021-11-10
Add checkpoints to staking contract

cygnusv:
Adds checkpoints to staking contract, so that stakers can collect and delegate voting power in proportion to their stake size.
Some important aspects to note:
* Delegation of stake voting power is controlled by the owner (i.e., the owner decides to whom delegate their vote)
* Threshold model of staking roles allows for different stakes, associated to different operators, but with the same stake owner. Conversely, this means that the same owner address can have multiple stakes associated to different operators. In practice, in this situation the owner must delegate each stake individually.
* Because of the particularities of the Threshold model of staking roles, the delegation function requires an extra parameter, that is, the operator address, in order to uniquely identify each stake:
```js
    /// @notice Delegate voting power from the stake associated to the
    ///         `operator` to a `delegatee` address. Caller must be the owner
    ///         of this stake.
    function delegateVoting(address operator, address delegatee) public {
        delegate(operator, delegatee);
    }
```

### DIFF
--- a/contracts/staking/TokenStaking.sol
+++ b/contracts/staking/TokenStaking.sol
@@ -251,6 +251,8 @@ contract TokenStaking is Ownable, IStaking, Checkpoints {
         /* solhint-disable-next-line not-rely-on-time */
         operatorStruct.startTStakingTimestamp = block.timestamp;
 
+        increaseStakeCheckpoint(operatorStruct.owner, amount);
+
         emit OperatorStaked(
             StakeType.T,
             msg.sender,
@@ -284,6 +286,9 @@ contract TokenStaking is Ownable, IStaking, Checkpoints {
         operatorStruct.beneficiary = keepStakingContract.beneficiaryOf(
             operator
         );
+
+        increaseStakeCheckpoint(operatorStruct.owner, tAmount);
+
         emit OperatorStaked(
             StakeType.KEEP,
             operatorStruct.owner,
@@ -325,6 +330,8 @@ contract TokenStaking is Ownable, IStaking, Checkpoints {
         operatorStruct.owner = msg.sender;
         operatorStruct.authorizer = authorizer;
         operatorStruct.beneficiary = beneficiary;
+        
+        increaseStakeCheckpoint(operatorStruct.owner, tAmount);
 
         emit OperatorStaked(
             StakeType.NU,
@@ -571,6 +578,7 @@ contract TokenStaking is Ownable, IStaking, Checkpoints {
         OperatorInfo storage operatorStruct = operators[operator];
         operatorStruct.tStake += amount;
         emit ToppedUp(operator, amount);
+        increaseStakeCheckpoint(operatorStruct.owner, amount);
         token.safeTransferFrom(msg.sender, address(this), amount);
     }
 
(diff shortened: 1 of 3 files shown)


contracts/staking/TokenStaking.sol
```diff
@@ -758,24 +775,32 @@ contract TokenStaking is Ownable, IStaking {
...
         emit TokensSeized(operator, slashedAmount, true);
         if (undelegatedAt != 0) {
             operatorStruct.keepInTStake = 0;
+            decreaseStakeCheckpoint(operator, oldKeepInTStake);
+        } else {
+            decreaseStakeCheckpoint(
+                operator,
+                oldKeepInTStake - operatorStruct.keepInTStake
+            );
         }
```
vzotova, line 804:
```suggestion
        }
        decreaseStakeCheckpoint(
            operator,
            oldKeepInTStake - operatorStruct.keepInTStake
        );
```
I think this will cost less bytes

contracts/staking/TokenStaking.sol
```diff
@@ -1021,6 +1051,13 @@ contract TokenStaking is Ownable, IStaking {
         return slashingQueue.length;
     }
 
+    /// @notice Delegate voting power from the stake associated to the
+    ///         `operator` to a `delegatee` address. Caller must be the owner
+    ///         of this stake.
+    function delegateVoting(address operator, address delegatee) public {
```
vzotova, line 1057: Can it be external?

vzotova (commented):
I guess there is missing commit with new optimizer runs or something causing the TokenStaking deployment to fail during tests

contracts/staking/TokenStaking.sol
cygnusv: Yes it can, good idea

pdyraga:
> I guess there is missing commit with new optimizer runs or something causing the TokenStaking deployment to fail during tests

Have we not just exceeded 24KB contract size limit with the additions from this PR?

cygnusv:
> > I guess there is missing commit with new optimizer runs or something causing the TokenStaking deployment to fail during tests
> 
> Have we not just exceeded 24KB contract size limit with the additions from this PR?

Yeah, but reducing the number of optimization rounds allows to be right under 24KB

contracts/staking/TokenStaking.sol
cygnusv: Done

vzotova (approved):

contracts/staking/TokenStaking.sol
```diff
@@ -697,7 +714,7 @@ contract TokenStaking is Ownable, IStaking {
             "Too much to unstake"
         );
         operatorStruct.nuInTStake -= amount;
-
+        decreaseStakeCheckpoint(operator, amount);
```
pdyraga, line 717: Nitpick but we could keep the same order of operations of emitting event, updating staked amount, and decreasing checkpoint in `unstake`, `unstakeKeep`, and `unstakeNu`.

contracts/staking/TokenStaking.sol
```diff
@@ -931,6 +960,13 @@ contract TokenStaking is Ownable, IStaking {
         }
     }
 
+    /// @notice Delegate voting power from the stake associated to the
+    ///         `operator` to a `delegatee` address. Caller must be the owner
+    ///         of this stake.
+    function delegateVoting(address operator, address delegatee) external {
+        delegate(operator, delegatee);
+    }
```
pdyraga, line 968: What if we rename `delegate` function in `T` token contract to `delegateVoting` to keep them aligned?

https://github.com/threshold-network/solidity-contracts/blob/fa83b98566b7c786a499fc438d97b4dd3eee0293/contracts/token/T.sol#L79

We could also rename the function in `Checkpoint` so that we do not need to have `function delegate(address operator, address delegatee)` a few lines below and just put this logic into `delegateVoting`.

test/staking/TokenStaking.test.js
```diff
@@ -5818,6 +6063,12 @@ describe("TokenStaking", () => {
           ])
         })
 
+        it("should decrease the delegatee voting power", async () => {
+          expect(await tokenStaking.getVotes(delegatee.address)).to.equal(
+            expectedAmount
+          )
+        })
```
pdyraga, line 6070: I think we should add a test that it `should decrease the total voting power` as well.

pdyraga (approved):
There is some infrastructure-level problem with the draft system tests job. Given system tests are still in draft, I am ignoring it and merging.

PR #27, opened 2021-11-11
Improvements to KeepStake owner resolver

pdyraga:
See #26 
See #7 
See https://github.com/keep-network/keep-core/pull/2697

Three improvements to `KeepStake` owner resolver here.

First, we pass legacy keep staking contract to `KeepStake` via the constructor.

Second, we now allow Governance to add new operator-owner mappings to `KeepStake`. In practice, this will be needed only if someone decides to stake their KEEP token grant in KEEP network after 2021-11-11 when the snapshot was taken. These functions should never be needed but it is good to have them just in case.

Last but not least, removed duplicate managed grant/grantee entries. Due to the bug in grantee resolution script, all managed grants were duplicated as standard grantees.  See keep-network/keep-core@1ad4b68


contracts/staking/KeepStake.sol
```diff
@@ -3,17 +3,83 @@
...
+        address owner = operatorToManagedGrant[operator];
+        if (owner != address(0)) {
+            return IManagedGrant(owner).grantee();
+        }
+        owner = operatorToGrantee[operator];
+        if (owner != address(0)) {
+            return owner;
+        }
+
         if (operator == 0x855A951162B1B93D70724484d5bdc9D00B56236B) {
```
vzotova, line 83: What do you think about add these operators to mappings in the constructor? This will be free from gas and contact size perspective and will make `resolveOwner` simpler

contracts/staking/KeepStake.sol
```diff
@@ -3,17 +3,83 @@
...
+    function setGrantee(address operator, address grantee) external onlyOwner {
+        operatorToGrantee[operator] = grantee;
+    }
+
+    /// @notice Resolves KEEP stake owner for the provided operator address.
+    function resolveOwner(address operator) external view returns (address) {
+        address owner = operatorToManagedGrant[operator];
+        if (owner != address(0)) {
+            return IManagedGrant(owner).grantee();
+        }
```
vzotova, line 77: Small question: I saw that ManagedGrant is not checking owner is zero address or not in the constructor. Is that possible that resolved address will be zero and what do you think about adding check before returning for whole method?

contracts/staking/KeepStake.sol
pdyraga: Hmm... I've left them here exactly for this reason. If we add them in the constructor, the deployment will be expensive, because of SSTOREs, right? Right now, it's ~2k to resolve the owner from what I checked.

contracts/staking/KeepStake.sol
pdyraga: It is possible in theory but not in practice. We'd either need to create a grant for 0x0 which is by itself a terrible mistake in the process, or we'd need to have the grantee requesting for reassignment to 0x0 and us (grant manager) accepting this reassignment.

contracts/staking/KeepStake.sol
vzotova: These values are stored in contract size, so it's also cost something, but not sure more or less than SSTORE.
Also I think resolving will read less, because will be only two reads from mapping instead of two reads + ifs (in case of some static value). Of course it's only guessing, need to measure to be sure. My main point that will be easier to review this method with simpler logic.
But this is not critical

contracts/staking/KeepStake.sol
vzotova: Ok, if you're sure then fine

contracts/staking/KeepStake.sol
pdyraga: I tried experimenting with gas costs but Remix refuses to deploy my contracts... I extracted snapshotted addresses to separate functions in https://github.com/threshold-network/solidity-contracts/pull/27/commits/6ada9c9b193bcb139a71f840c59fd8080bb8f1c5. They are internal so should be copied by the compiler anyway but at least reading the code should be easier now. We can revisit and compare costs once #22 is merged - hardhat deploy plugin displays the gas cost of this operation.

contracts/staking/KeepStake.sol
pdyraga: Actually... 

I am comfortable with not checking what is returned by `ManagedGrant.grantee()` but we need to check what is returned by the final `TokenStaking.ownerOf`. In https://github.com/threshold-network/solidity-contracts/pull/27/commits/e473cb064db4131deaed786a8111d1edc59cc22e I added a check to revert if `KeepStake.resolveOwner` does not know the operator and resolves the owner to 0x0.

Without reverting, the following corner case is possible:
1. Someone stakes liquid T in T TokenStaking contract.
2. They call `TokenStaking.refreshKeepStakeOwner` for their operator.
3. Owner resolves to 0x0 because legacy staking contract does not know that operator.
4. Owner is updated to 0x0.
5. Sad face.

Ideally, we'd revert in `TokenStaking` contract but we are already concerned about its size and we need to add more functionalities there, such as disabled applications. If, after implementing disabled applications, we can move that check there, that would be ideal.

vzotova: I agree that proper place for `require` is in `TokenStaking` but free size is not enough for `disableApp` right now, so maybe after next iteration of optimizations.
Also would be great to have test that checks case `owner == address(0)`

pdyraga: Oops, my fault. Please see https://github.com/threshold-network/solidity-contracts/pull/27/commits/1c346e07c79a46c3ce645458e366a29a3f6438ed.

test/staking/KeepStake.test.js
```diff
@@ -0,0 +1,152 @@
...
+        )
+      })
+
+      it("should fallback to Keep staking contract", async () => {
+        expect(await keepStake.resolveOwner(operator)).to.equal(owner)
+      })
+
+      it("should revert if Keep staking does not know the operator", async () => {
+        await expect(
+          await keepStake.resolveOwner(authorizer)
```
vzotova, line 101:
```suggestion
          keepStake.resolveOwner(authorizer)
```
probably test fails because of that

pdyraga: facepalm

vzotova (approved):

test/staking/KeepStake.test.js
pdyraga: Fixed, rebased, force pushed! Build is green now.

PR #29, opened 2021-11-15
System tests infrastructure

lukasz-zimnoch:
Here we introduce system tests infrastructure and a basic test scenario for staking. System tests use Hardhat mainnet forking feature and can be triggered using:
```
FORKING_URL=<forking-url> yarn test:system
```
where `forking-url` should point to a HTTPS endoint of an Ethereum node with access to archival data (Alchemy recommended). Additionally, system tests are triggered via GitHub Actions upon each pull request if `contracts/staking` or `test/system` directories were modified.


.github/workflows/contracts.yaml
```diff
@@ -8,6 +8,23 @@ on:
   workflow_dispatch:
 
 jobs:
```
pdyraga, line 10: @michalinacienciala Could you please review the change here?

contracts/staking/ILegacyTokenStaking.sol
```diff
@@ -19,6 +19,18 @@ interface IKeepTokenStaking {
...
+    /// Contracts using delegated authority
+    /// cannot be authorized with `authorizeOperatorContract`.
+    /// Instead, authorize `getAuthoritySource(_operatorContract)`.
+    /// @param operator address of stake operator.
+    /// @param operatorContract address of operator contract.
+    function authorizeOperatorContract(
+        address operator,
+        address operatorContract
+    ) external;
+
```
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/threshold-network/solidity-contracts/pull/29/commits/deab9e7e71fcae1ac48bd1e3f66617c9e34b6658

test/helpers/contract-test-helpers.js
```diff
@@ -25,12 +25,55 @@ async function mineBlock() {
...
+    await purseSigner.sendTransaction({
+      to: accountAddress,
+      value: ethers.utils.parseEther(amount || "1"),
+    })
+  }
+
+  // Return the account's signer.
+  return await ethers.getSigner(accountAddress)
+}
+
```
pdyraga, line 68: Non blocking but I think it would be worth looking if we can extract it to https://github.com/keep-network/hardhat-helpers and then import here.

test/system/init-contracts.js
```diff
@@ -0,0 +1,158 @@
...
+async function initContracts() {
+  const deployer = await ethers.getSigner(0)
+
+  const keepToken = await resolveKeepToken()
+  const nuCypherToken = await resolveNuCypherToken()
+  const keepTokenStaking = await resolveKeepTokenStaking()
+  const nuCypherStakingEscrow = await resolveNuCypherStakingEscrow()
+  const keepRegistry = await resolveKeepRegistry()
+
+  const tToken = await deployTToken(to1e18(1000000))
```
pdyraga, line 22: Should be `10 000 000 000`. Non blocking but we should update it at some point to use real-world value.

test/system/init-contracts.js
```diff
@@ -0,0 +1,158 @@
...
+
+  const keepVendingMachine = await deployVendingMachine(
+    keepToken,
+    tToken,
+    to1e18(1000000),
+    to1e18(500000)
```
pdyraga, line 28: 45% of the supply here please, same for `nuCypherVandingMachine`. Non-blocking one.

.github/workflows/contracts.yaml
michalinacienciala: Looking good!

pdyraga (approved):

PR #30, opened 2021-11-15
Ability to disable application

### LINKED ISSUE #28: `Disabled` state of application in `TokenStaking`
Comment by @pdyraga from #7 

> Separately, from the paused state, we need something to let stakers exit the application in case some critical vulnerability was found without giving the application a chance to slash them again.
>
> I think the state transition should be: active <-> paused -> disabled.
>
> We could rename disabled to paused and have one more function callable only by the contract owner (governance/DAO) that would set real ApplicationInfo.disabled boolean to true. For disabled application, stakers could call some function (forceDecreaseAuthorization?) to decrease their authorization for that application to 0 without having to go through the two-step decrease process.

vzotova:
Closes #28

Added new `disabled` status for application, method `disableApplication` and method `forceDecreaseAuthorization`  to nullify authorized amount to disabled application.
Also PR includes code size optimizations (current size is 24.234) and additional tests.

### DIFF
--- a/contracts/staking/TokenStaking.sol
+++ b/contracts/staking/TokenStaking.sol
@@ -1532,10 +1532,8 @@ contract TokenStaking is Ownable, IStaking, Checkpoints {
             }
         }
 
-        address[] storage authorizedApplications = operatorStruct
-            .authorizedApplications;
-        assembly {
-            sstore(authorizedApplications.slot, newLength)
+        for (index = newLength; index < length; index++) {
+            operatorStruct.authorizedApplications.pop();
         }
     }
 
--- a/contracts/staking/TokenStaking.sol
+++ b/contracts/staking/TokenStaking.sol
@@ -29,6 +29,13 @@ contract TokenStaking is Ownable, IStaking, Checkpoints {
     using PercentUtils for uint256;
     using SafeCast for uint256;
 
+    enum ApplicationStatus {
+        NOT_APPROVED,
+        APPROVED,
+        PAUSED,
+        DISABLED
+    }
+
     struct OperatorInfo {
         uint96 nuInTStake;
         address owner;
@@ -47,8 +54,7 @@ contract TokenStaking is Ownable, IStaking, Checkpoints {
     }
 
     struct ApplicationInfo {
-        bool approved;
-        bool paused;
+        ApplicationStatus status;
         address panicButton;
     }
 
@@ -96,7 +102,10 @@ contract TokenStaking is Ownable, IStaking, Checkpoints {
         uint96 amount
     );
     event MinimumStakeAmountSet(uint96 amount);
-    event ApplicationApproved(address indexed application);
+    event ApplicationStatusChanged(
+        address indexed application,
+        ApplicationStatus indexed newStatus
+    );
     event AuthorizationIncreased(
         address indexed operator,
         address indexed application,
@@ -118,7 +127,6 @@ contract TokenStaking is Ownable, IStaking, Checkpoints {
         uint96 amount,
         bool indexed successfulCall
     );
-    event ApplicationPaused(address indexed application);
     event PanicButtonSet(
         address indexed application,
         address indexed panicButton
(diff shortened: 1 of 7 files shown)


pdyraga:
@vzotova I've just merged #25, can you please do rebase magic here? This will help me with the review.

vzotova:
> @vzotova I've just merged #25, can you please do rebase magic here? This will help me with the review.

👍 Done

contracts/staking/TokenStaking.sol
```diff
@@ -269,7 +277,7 @@ contract TokenStaking is Ownable, IStaking, Checkpoints {
     ///         stake amount from KEEP staking contract. Can be called by
     ///         anyone.
     function stakeKeep(address operator) external override {
-        require(operator != address(0), "Operator must be specified");
+        require(operator != address(0), "Parameters must be specified");
```
pdyraga, line 280: Is it more efficient for the contract size to have the same string everywhere? 🤔

vzotova: Yep, now it's stored only once, it's the most efficient optimization here 😄

contracts/staking/TokenStaking.sol
```diff
@@ -1463,8 +1531,11 @@ contract TokenStaking is Ownable, IStaking, Checkpoints {
...
             }
         }
-        for (index = newLength; index < length; index++) {
-            operatorStruct.authorizedApplications.pop();
+
+        address[] storage authorizedApplications = operatorStruct
+            .authorizedApplications;
+        assembly {
+            sstore(authorizedApplications.slot, newLength)
         }
```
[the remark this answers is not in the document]
vzotova: I don't think it's explained somewhere. You're right, no real cleaning, but in case of adding new element it will rewrite anyway. If you think it's too hacky - I'll change it back

test/staking/TokenStaking.test.js
```diff
@@ -2027,6 +2102,148 @@ describe("TokenStaking", () => {
...
+            .connect(staker)
+            .forceDecreaseAuthorization(
+              operator.address,
+              application1Mock.address
+            )
+        ).to.be.revertedWith("Application is not disabled")
+      })
+    })
+
+    context("when application is not authorized", () => {
```
pdyraga, line 2157: Should we wrap and the next context into another one saying "when application is disabled"? Something like:
```
context("when application is disabled", () => {
  context("when application was not authorized", () => {
  }

  context("when application was authorized", () => {
  }
})
```

test/staking/TokenStaking.test.js
```diff
@@ -6467,4 +6856,134 @@ describe("TokenStaking", () => {
...
+            application2Mock.address,
+            amount
+          )
+          await extendedTokenStaking.cleanAuthorizedApplications(
+            operator.address,
+            2
+          )
+        })
+
+        it("should remove only first application", async () => {
```
pdyraga, line 6979: Is this description accurate? It removed two applications.

contracts/staking/TokenStaking.sol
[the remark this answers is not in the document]
pdyraga: I see it for the first time and I feel a little itchy about it 😆 We should not have that many applications in that array so if you ask me, I feel more comfortable with the previous version.

test/staking/TokenStaking.test.js
vzotova: Fixed
✔️

contracts/staking/TokenStaking.sol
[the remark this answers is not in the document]
vzotova: Ok, no problem, changed back. We can consider such thing as emergency way to get 90 bytes more 
✔️

test/staking/TokenStaking.test.js
pdyraga: My suggestion was based on the fact I was confused this application is disabled *and* not authorized. I was about to write a comment it is not possible to disable unapproved application (I missed approval with authorization). 

I think `context` is not only about extracting common setup but also about... wrapping elements into some additional information context. But if we want to avoid it, maybe we can provide more info in test names:
- when application was not authorized and god disabled
- when application was authorized and got disabled
? 

Both options are equally good for me.

vzotova: good point, renamed
✔️

pdyraga (approved):

PR #32, opened 2021-11-19
Use @keep-network/hardhat-helpers package

nkuba:
`@keep-network/hardhat-helpers` package defines common helper functions that can be used in hardhat deployment and test scripts.


r-czajkowski (approved):
LGTM!

cygnusv (approved):

PR #34, opened 2021-11-29
Deploy `VendingMachine` contract for NU token

r-czajkowski:
This PR:

* deploys the mocked NU token expect the main network. For mainnet script saves the NU token artifact from `external/mainnet/NuCypherToken.json` in the `deployments/<network>` dir,
* deploys `VendingMachine` contract for NU token,
* add `prepare-arrtifacts.sh` script- hardhat by default saves deployment artifacts to `deployments/<network>`since we're publishing separate package for each network we can remove one level from this directory structure so we want to save them in `artifacts` dir. This script should be run before publishing to the npm registry as well.


deploy/04_deploy_vending_machine.ts
```diff
@@ -0,0 +1,75 @@
...
+    const totalSupply = await read(tokenArtifactName, "totalSupply")
+    const vendingMachine = await deployments.deploy(
+      vendingMachineArtifactName,
+      {
+        contract: "VendingMachine",
+        from: deployer,
+        args: [
+          Token.address,
+          T.address,
+          totalSupply,
```
cygnusv, line 53: For NuCypher's VendingMachine deployment, we can't use the token `totalSupply()`, as it will report an incorrect figure for vending machine purposes. The correct number is the `currentPeriodSupply` in NuCypher's `StakingEscrow` contract.

deploy/01_resolve_nucypher_token.ts
```diff
@@ -0,0 +1,44 @@
...
+    hre.network.name === "mainnet" &&
+    NuCypherToken &&
+    helpers.address.isValid(NuCypherToken.address)
+  ) {
+    log(`using external NuCypherToken at ${NuCypherToken.address}`)
+    await deployments.save("NuCypherToken", NuCypherToken)
+  } else {
+    log(`deploying NuCypherToken stub`)
+
+    const NU_SUPPLY = to1e18("1000000000") // 1B NU
```
cygnusv, line 23: This number should be changed if we want to make a testnet deployment that is closer to the real setting. We deployed the token with a 3.8 B supply, but most of it is on escrow in the `StakingEscrow` contract. The actual issued supply to date is around 1.34 B, and it will still be around that number when the merge happens.

deploy/04_deploy_vending_machine.ts
```diff
@@ -0,0 +1,75 @@
...
+  const { deployer } = await getNamedAccounts()
+  const { read } = deployments
+
+  const T = await deployments.get("T")
+  const tTotalSupply = await read("T", "totalSupply")
+
+  // We're wrapping 100% of the minted KEEP and will be allocating 45% of the
+  // minted T tokens. The remaining T tokens will be in the future distributed
+  // between another instance of the VendingMachine (which will be wrapping NU
+  // token) and a DAO treasury.
```
michalinacienciala, line 22: Could you update the comment? It got outdated after the recent changes.

deploy/05_transfer_t.ts
```diff
@@ -8,6 +8,15 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
...
+
+  const vendinMachines = [
+    { tokenSymbol: "KEEP", vendingMachineAddress: VendingMachineKeep.address },
+    {
+      tokenSymbol: "NU",
+      vendingMachineAddress: VendingMachineNuCypher.address,
+    },
+  ]
 
   // There will be 10B T minted on the production environment. The 45% of this
```
michalinacienciala, line 21: Could you update the comment? The first two sentences are ok, but rest can be removed after you've added vending machine for NU.

michalinacienciala (changes requested):
I left a couple of comments. I haven't tested ropsten deployment yet (I'm getting some errors likely related to my local config), I'm working on it, but I'm posting what I have found so far. 

Also,  one more thing.... Could you change the printed name of the token from `KEEP` to `T` in the line [18](https://github.com/threshold-network/solidity-contracts/blob/32d5689e1e3c101c74f09102a5c5173d6aac7969/deploy/02_mint_t.ts#L18) of `02_mint_t.ts`?
That's something I overlooked when I was working on the file and we could fix this as part of this PR.

scripts/prepare-artifacts.sh
```diff
@@ -0,0 +1,54 @@
...
+
+[ ! -d "${SOURCE_DEPLOYMENT_DIR}" ] && {
+  echo "$SOURCE_DEPLOYMENT_DIR does not exist" >&2
+  exit 1
+}
+
+rm -rf $DESTINATION_ARTIFACTS_DIR
+cp -r deployments/$NETWORK $DESTINATION_ARTIFACTS_DIR
+
+echo "Deployment artifacts copied to: $DESTINATION_ARTIFACTS_DIR"
```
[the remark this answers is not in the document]
r-czajkowski: Sure I will add! I skipped it on purpose because I thought you might want to add this or develop another solution on your own.

deploy/04_deploy_vending_machine.ts
r-czajkowski: Thanks for flagging this! Please see 5758b0b and 180ca7d.

deploy/01_resolve_nucypher_token.ts
r-czajkowski: Good point! Please see 8096a55

r-czajkowski:
@michalinacienciala and @cygnusv PR is ready for the next round.

scripts/prepare-artifacts.sh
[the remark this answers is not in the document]
r-czajkowski: Added in 6e4df13.

deploy/05_transfer_t.ts
r-czajkowski: Please see d1800a2

michalinacienciala (commented):
I've tested deployment scripts on `ropsten` environment and noticed that there is a problem with Tenderly verification of `VendingMachineKeep` and `VendingMachineNuCypher` contracts - see step "Deploy contracts" here: https://github.com/threshold-network/solidity-contracts/runs/4369450160?check_suite_focus=true (the workflow was run on a clone of `deploy-vending-machine-for-nu` branch, with one change that switches the `--dry-run` on for the step publishing NPM package).
The error message is quite vague, so I'm not sure what's the root cause (will try to investigate tomorrow). Good news is that it is not blocking the deployment, so I think we can merge the changes (if approved by the reviewers) and work on fixing the Tenderly problems as a separate task (we're gonna need to add this task to Jira if we decide to leave it for later).

I also tested the deployment on `hardhat` network, looking good to me.

I don't feel fully confident about my review of `deploy/04_deploy_vending_machine.ts` as my experience with TS is quite limited, so I would appreciate other reviewers paying increased attention to that file.

cygnusv (approved):
LGTM!

deploy/01_resolve_nucypher_token.ts
```diff
@@ -0,0 +1,49 @@
...
+      log: true,
+    })
+
+    await execute(
+      "NuCypherToken",
+      { from: deployer },
+      "mint",
+      deployer,
+      NU_SUPPLY
+    )
```
cygnusv, line 41: I missed something here on my review: the NuCypherToken mints the supply at construction time and doesn't have a `mint()` function to use later, so what you should do here is pass `NU_SUPPLY` to the constructor. The whole supply would be now the balance of the deployer account.

cygnusv (changes requested):
Just realized a small detail about the NU token stub deployment.

deploy/01_resolve_nucypher_token.ts
r-czajkowski: I used an existing [`TestToken`](https://github.com/threshold-network/solidity-contracts/blob/main/contracts/test/TestToken.sol) contract to deploy stub NU token that's why we mint tokens this way. So should I create a new contract that better matches the real NuCypher token implementation or does it work for you for stub deployment?

cygnusv: Oh I see. Nevermind then, for the purposes of preliminary integrations that's fine. The NuCypher token doesn't have anything out of the ordinary for an ERC20 token anyway.

cygnusv (approved):

deploy/04_deploy_vending_machine.ts
```diff
@@ -0,0 +1,119 @@
+import { HardhatRuntimeEnvironment } from "hardhat/types"
```
nkuba, line 1: The solution proposed in this script seems very sophisticated. I'm afraid it might be too complicated. What if we go with a much simpler script, consistent with the other ones, and just enhance the previous version https://github.com/threshold-network/solidity-contracts/blob/32d5689e1e3c101c74f09102a5c5173d6aac7969/deploy/03_deploy_vending_machine.ts8 by adding a few lines with a deployment of another `VendingMachine` for NU:
```ts
 const T_ALLOCATION_NU = tTotalSupply.mul(45).div(100)

  const vendingMachine = await deployments.deploy("VendingMachineNuCypher", {
    contract: "VendingMachine",
    from: deployer,
    args: [NuCypherToken.address, T.address, nuTotalSupply, T_ALLOCATION_NU],
    log: true,
  })
```

IMHO the deployment scripts should be kept simple and easily maintainable. There will be just two types of vending machines, so we don't have to think about being too universal.

nkuba (changes requested):

deploy/04_deploy_vending_machine.ts
r-czajkowski: Yeah, I see your point. Should we split scripts into two files eg.`04_deploy_vending_machine_keep` and `05_deploy_vending_machine_nu`? What do you think?

nkuba: I don't have a strong opinion about splitting the script. I can see the advantages of both solutions. Up to you!

r-czajkowski: Please see cab0c53 and 03416c4.

michalinacienciala (approved):

michalinacienciala (approved):
Tested the scripts on `hardhat` network and on `ropsten` https://github.com/threshold-network/solidity-contracts/runs/4445794511?check_suite_focus=true).
We still can see the problems with verification in Tenderly for contracts `VendingMachineKeep` and `VendingMachineNuCypher` (not a blocker), but everything else looks good.

deploy/06_transfer_t.ts
```diff
@@ -8,29 +8,36 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
...
+    await execute(
+      "T",
+      { from: deployer },
+      "transfer",
+      vendingMachineAddress,
+      T_TO_TRANSFER
+    )
+
+    console.log(
+      `transfered ${from1e18(
```
nkuba, line 36:
```suggestion
      `transferred ${from1e18(
```

nkuba (dismissed):

michalinacienciala (approved):

PR #41, opened 2021-12-13
SWAG all the things

pdyraga:
Turn on the SWAG on Etherscan. Example: https://etherscan.io/address/0x27321f84704a599ab740281e285cc4463d89a3d5#code

image


contracts/governance/Checkpoints.sol
```diff
@@ -1,4 +1,20 @@
...
+// ██████████████     ▐████▌     ██████████████
+//               ▐████▌    ▐████▌
+//               ▐████▌    ▐████▌
+//               ▐████▌    ▐████▌
+//               ▐████▌    ▐████▌
+//               ▐████▌    ▐████▌
+//               ▐████▌    ▐████▌
+//
+//
+//           Trust math, not hardware.
```
cygnusv, line 17: Let's find here a common motto. We can use the Threshold branding material.

pdyraga: Hmmm... I was sure this is what we ended up with. Tagging @dougvk and @willthesis here.

dougvk: @cygnusv the branding material isn't for our code. do you have another suggestion? we can discuss on discord if needed

pdyraga: Removed the motto, we can always add it later again.

contracts/governance/GovernorParameters.sol
```diff
@@ -1,4 +1,17 @@
-// SPDX-License-Identifier: MIT
+// SPDX-License-Identifier: GPL-3.0-or-later
```
pdyraga, line 1: Flagging this change @cygnusv - did we have MIT there on purpose? I was sure we are going with GPL everywhere?

cygnusv: That's a mistake 🙈

pdyraga: So it's fixed now :)

cygnusv (approved):
SWAG it!

PR #42, opened 2021-12-14
Tag testnet deployment

michalinacienciala:
When building and publishing token dashboard on testnet we need to use
dependency to the latest `threshold-network/solidity-contracts` package
containing contracts deployed on that testnet. In order to do that
easily, we should use a tag for all such `solidity-contracts` packages.
Here we're adding the tagging. NPM publish step is executed only for
`workflow_dispatch` event, that means that we can use `environment`
input for tagging the package.


r-czajkowski (approved):
LGTM!

PR #48, opened 2021-12-17
Adds time restriction to unstaking NU

vzotova:
Adds restriction to `unstakeNu`/`unstakeAll`, same as in `unstakeT`.
Besides that softened restrictions for Keep stakers: no need to check timestamp because operator address protected in Keep staking contract.

### DIFF
--- a/contracts/staking/TokenStaking.sol
+++ b/contracts/staking/TokenStaking.sol
@@ -61,7 +61,7 @@ contract TokenStaking is Initializable, IStaking, Checkpoints {
         address authorizer;
         mapping(address => AppAuthorization) authorizations;
         address[] authorizedApplications;
-        uint256 startTStakingTimestamp;
+        uint256 startStakingTimestamp;
     }
 
     struct AppAuthorization {
@@ -286,7 +286,7 @@ contract TokenStaking is Initializable, IStaking, Checkpoints {
 
         operatorStruct.tStake = amount;
         /* solhint-disable-next-line not-rely-on-time */
-        operatorStruct.startTStakingTimestamp = block.timestamp;
+        operatorStruct.startStakingTimestamp = block.timestamp;
 
         increaseStakeCheckpoint(operator, amount);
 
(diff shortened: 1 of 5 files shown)


contracts/staking/TokenStaking.sol
```diff
@@ -734,8 +736,10 @@ contract TokenStaking is Ownable, IStaking, Checkpoints {
         );
         operatorStruct.tStake -= amount;
         require(
-            operatorStruct.tStake >= minTStakeAmount ||
-                operatorStruct.startTStakingTimestamp + MIN_STAKE_TIME <=
+            operatorStruct.nuInTStake != 0 ||
+                operatorStruct.keepInTStake != 0 ||
```
cygnusv, line 740: Does this mean that it's possible to satisfy the entire require by leaving a small amount of NU/KEEP?

vzotova: yep, because staker can change operator only if there is no more staked nu, so 10^-18 Nu will prevent from using other operator

pdyraga:
@vzotova Do we still need this change given #66 ?

vzotova:
> @vzotova Do we still need this change given #66 ?

yep, it will protect from fast changing of staking provider for nu legacy stakers

cygnusv (approved):

contracts/staking/TokenStaking.sol
```diff
@@ -1114,17 +1128,17 @@ contract TokenStaking is Initializable, IStaking, Checkpoints {
...
 
-    /// @notice Returns start staking timestamp for T stake.
+    /// @notice Returns start staking timestamp for T/NU stake.
     /// @dev    This value is set at most once, and only when a stake is created
-    ///         with T tokens. If a stake is created from a legacy stake,
-    ///         this value will remain as zero
-    function getStartTStakingTimestamp(address operator)
+    ///         with T or NU tokens. If a stake is created from a legacy KEEP
+    ///         stake, this value will remain as zero
+    function getStartStakingTimestamp(address operator)
```
pdyraga, line 1135: Why don't we set this value for `stakeKeep` as well? 

We do not have to use it for undelegation validation given it does not make sense for KEEP stakers but if `getStartStakingTimestamp` works only for liquid T and NU, the function name is confusing. 

We could also use `getStartStakingTimestamp` in the dApp to display when someone started staking and if they staked, without having to scan events.

vzotova: Yes, we could do that. Just tried to minimize changes

pdyraga: See #73

PR #54, opened 2021-12-23
Add scripts directory to NPM package

nkuba:
We need to include the scripts directory as it contains files that are
executed as part of NPM scripts, i.e. `postinstall`.

Another change in this PR is the range for `@keep-network/keep-core` dependency. We should have a range specific for development packages. The previous solution was pulling packages for ropsten.


r-czajkowski (approved):
LGTM!

PR #56, opened 2021-12-28
Fix postinstall script

nkuba:
This PR fixes usage of postinstall script in other projects, where this package is a dependency.


r-czajkowski (approved):
LGTM!

PR #71, opened 2022-01-25
Staking provider: follow-up

### LINKED ISSUE #66: Per-application operators
# Overview
This issue presents the idea we discussed with @vzotova and @cygnusv to enable per-application operators in T staking contract.

# In Depth

## Current Functionality

KEEP network uses the concept of an operator as an address of a party authorized to operate in the network on behalf of a given owner. The operator handles the everyday operations on the delegated stake without actually owning the staked tokens. An operator can not simply transfer away delegated tokens, however, it should be noted that the operator’s misbehavior may result in slashing tokens and thus the entire staked amount is indeed at stake. An operator is a hot wallet whose private key is configured in ECDSA and random beacon node configuration. The operator's Ethereum address uniquely identifies the node on-chain and off-chain. P2P network identifier is derived from the operator's public key. Operator address is immutable for the given stake and must be unique across the network.

NuCypher network does not have a concept of an operator but does have a concept of a worker that is pretty much the same. One major difference is that the worker can be easily changed for the given stake delegation, while for KEEP the operator is immutable. Worker's Ethereum address uniquely identifies PRE node on-chain and PRE node is configured with worker's private key.

KEEP staking contract assumes the same operator is used for the given stake delegation in all applications. That is, it is impossible to use another operator for random beacon and another operator for ECDSA in the same delegation. The owner must create two separate delegations and have twice as many tokens to use per-application unique operators.

This setup has been ported to the first implementation of T staking contract - the operator is unique in the network, it is immutable, and can not be changed for the given stake delegation. In retrospect, we identified this setup as sub-optimal given that there is no reason for PRE to restrict operator address immutability. Also, future applications implemented for T network may require a unique per-application operator address, not used anywhere else.

## Proposed Change

To enable per-application operators, the following changes are proposed:
- In T staking contract, _operator_ will be renamed to _staking provider_.
- _staking provider_ address will have to be unique.
- Application contracts (PRE, random beacon, ECDSA) will have to provide a mapping between _staking provider_ and _operator_.
- Off-chain clients will use _operator_ address for on-chain and off-chain identification.
- PRE node will use a concept of an _operator_ instead of a _worker_.
- It is up to the application to decide if _operator_ can be updated or not; most probably, PRE will support it but random beacon and ECDSA (tBTC) will not.
- Legacy KEEP stakers will continue to operate tBTC v1 off-chain nodes using their KEEP _operator_ address. This address will become _staking provider_'s address in T network.
- All T network off-chain clients will have to expose CLI-level feature to bind _staking provider_ with _operator_ address. For legacy KEEP stakers, _staking provider_ address (legacy KEEP _operator_ address) should not leave a VM and should not be imported into MetaMask.
- Legacy NU stakers and native T stakers - precisely their staking providers - are free to choose if the _staking provider_ should be a cold wallet or hot wallet.
- Legacy KEEP stakers - precisely their staking providers - will use KEEP _operator_ address (hot wallet) as a _staking provider_'s address in T network.
- For self-operating stakers, _staking provider_ address will be the one they control.
- `unstakeAll`, `unstakeT`, `unstakeKeep` will becone `onlyOwnerOrStakingProvider` in T staking contract.
- `topUp`, `topUpKeep`, `topUpNu` may be `onlyOwner` or `onlyOwnerOrStakingProvider` - still to decide.

### Legacy KEEP staker flow

In KEEP token dashboard:
- The authorizer has two transactions to execute: authorize T staking contract, stake into T staking contract.
- There is information saying "set up your PRE node" with a link to PRE documentation.

### Native T staker flow

In T token dashboard:
- Staker selects authorizer, beneficiary, and staking provider. All addresses defaults to staker's address.
- Staker has two transactions to execute: approve T transfer + stake into T staking contract.
- There is information saying "set up your PRE node" with a link to PRE documentation.

pdyraga:
Closes #66 

This PR changes no logic in tests or in the staking contract and is a followup of my comments to #67.

The goal is to use *staking provider* consistently instead of just *provider* everywhere.

It is also briefly mentioning the operator role in the RFC and in contract docs.

Last but not least, added `@dev` notes to some staking contract functions explaining legacy KEEP staking contract operator is staking provider in T staking contract.

### DIFF
--- a/contracts/staking/IStaking.sol
+++ b/contracts/staking/IStaking.sol
@@ -55,6 +55,8 @@ interface IStaking {
     ///         staking contract. No tokens are transferred. Caches the active
     ///         stake amount from KEEP staking contract. Can be called by
     ///         anyone.
+    /// @dev The staking provider in T staking contract is the legacy KEEP
+    ///      staking contract operator.
     function stakeKeep(address stakingProvider) external;
 
     /// @notice Copies delegation from the legacy NU staking contract to T
@@ -69,6 +71,8 @@ interface IStaking {
     ) external;
 
     /// @notice Refresh Keep stake owner. Can be called only by the old owner.
+    /// @dev The staking provider in T staking contract is the legacy KEEP
+    ///      staking contract operator.
     function refreshKeepStakeOwner(address stakingProvider) external;
 
     /// @notice Allows the Governance to set the minimum required stake amount.
--- a/contracts/staking/TokenStaking.sol
+++ b/contracts/staking/TokenStaking.sol
@@ -310,6 +310,8 @@ contract TokenStaking is Initializable, IStaking, Checkpoints {
     ///         staking contract. No tokens are transferred. Caches the active
     ///         stake amount from KEEP staking contract. Can be called by
     ///         anyone.
+    /// @dev The staking provider in T staking contract is the legacy KEEP
+    ///      staking contract operator.
     function stakeKeep(address stakingProvider) external override {
         require(stakingProvider != address(0), "Parameters must be specified");
         StakingProviderInfo storage stakingProviderStruct = stakingProviders[
@@ -394,6 +396,8 @@ contract TokenStaking is Initializable, IStaking, Checkpoints {
     }
 
     /// @notice Refresh Keep stake owner. Can be called only by the old owner.
+    /// @dev The staking provider in T staking contract is the legacy KEEP
+    ///      staking contract operator.
     function refreshKeepStakeOwner(address stakingProvider) external override {
         StakingProviderInfo storage stakingProviderStruct = stakingProviders[
             stakingProvider


docs/rfc-1-staking-contract.adoc
```diff
@@ -421,20 +431,22 @@ stake for that provider or if amount is higher than the cached legacy stake
...
+on the liquid T staked. Can be called only by the delegation owner or the
+staking provider.
 
 ==== `unstakeAll(address stakingProvider) external onlyOwnerOrStakingProvider(stakingProvider)`
 
 Sets cached legacy stake amount to 0, sets the liquid T stake amount to 0 and
 withdraws all liquid T from the stake to the owner. Reverts if there is at least one
-non-zero authorization. Can be called only by the delegation owner and provider.
+non-zero authorization. Can be called only by the delegation owner or the
+staking  provider.
```
vzotova, line 442:
```suggestion
non-zero authorization. Can be called only by the delegation owner or the
staking provider.
```

vzotova (approved):

docs/rfc-1-staking-contract.adoc
pdyraga: Addressed in #72 @ https://github.com/threshold-network/solidity-contracts/pull/72/commits/872717194a6c70898da7eb1b14db8b6700f35e5f

PR #72, opened 2022-01-25
Renamed ProviderStaked event to Staked

pdyraga:
~~Depends on #71~~

~~Leaving as a draft until #71 is merged.~~

It is not the staking provider who is staking in the staking contract.

The token owner is staking into the staking contract appointing the
staking provider as the one who will operate in the network on the
token owner's behalf.

PR #73, opened 2022-01-25
Write startStakingTimestamp in stakeKeep function, require 24h to pass to unstake

pdyraga:
~~Depends on #72~~ 
~~Leaving as a draft until #72  is merged.~~

We now Write `startStakingTimestamp` in `stakeKeep` function. This lets us capture the timestamp no matter the stake type. See #48 and [this comment](https://github.com/threshold-network/solidity-contracts/pull/48/files#r792011613) in particular.

Also, we do not allow to unstake before 24h passed since the stake has been delegated, no matter the stake type and no matter if there was a top-up or not. This requirement is stricter but allows us to avoid complicated unstake conditions and is less error-prone.

Simple rules, same for everyone.

### DIFF
--- a/test/staking/TokenStaking.test.js
+++ b/test/staking/TokenStaking.test.js
@@ -1,7 +1,7 @@
 const { expect } = require("chai")
 
 const { helpers } = require("hardhat")
-const { lastBlockTime, mineBlocks } = helpers.time
+const { lastBlockTime, mineBlocks, increaseTime } = helpers.time
 const { to1e18, to1ePrecision } = helpers.number
 
 const { AddressZero, Zero } = ethers.constants
@@ -2848,6 +2848,7 @@ describe("TokenStaking", () => {
     context("when staking provider unstaked T previously", () => {
       const amount = initialStakerBalance
       let tx
+      let blockTimestamp
 
       beforeEach(async () => {
         await tToken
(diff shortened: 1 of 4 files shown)


pdyraga:
Test will fail, working on it.

test/staking/TokenStaking.test.js
```diff
@@ -4542,9 +4337,6 @@ describe("TokenStaking", () => {
             authorizer.address
           )
 
-        await tToken.connect(staker).approve(tokenStaking.address, 1)
-        await tokenStaking.connect(staker).topUp(stakingProvider.address, 1)
-
```
vzotova, line 4547: What do you think about add something like that:
```         
await tokenStaking.connect(deployer).setMinimumStakeAmount(1)
await tToken.connect(staker).approve(tokenStaking.address, 1)
await tokenStaking.connect(staker).topUp(stakingProvider.address, 1)
```
To show T stake won't help to unstake 
Not necessary of course

vzotova (approved):
Looks good to me 
Some minor things related to tests, feel free to ignore them:

cygnusv (approved):

pdyraga:
@cygnusv @vzotova I have the two changes in unit tests @vzotova asked about almost ready but I'd suggest merging this PR given we have two approvals and I'll open another one with just test improvements within an hour. This should let us unblock the work.

test/staking/TokenStaking.test.js
pdyraga: Makes sense. Addressed in https://github.com/threshold-network/solidity-contracts/pull/74.

PR #78, opened 2022-01-31
Configure `KeepRegistry`

r-czajkowski:
Add the new deployment script that approves the `TTokenStaking` contract
in `KeepRegistry`. Each operator contract must be approved before it can
be authorized by a staker or used by a service contract.

### DIFF
--- a/contracts/test/KeepRegistryStub.sol
+++ b/contracts/test/KeepRegistryStub.sol
@@ -1,0 +1,19 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+pragma solidity 0.8.9;
+
+import "./IKeepRegistry.sol";
+
+contract KeepRegistryStub is IKeepRegistry {
+    event OperatorContractApproved(address operatorContract);
+
+    address public registryKeeper;
+
+    constructor() public {
+        registryKeeper = msg.sender;
+    }
+
+    function approveOperatorContract(address operatorContract) external {
+        emit OperatorContractApproved(operatorContract);
+    }
+}
--- a/contracts/test/KeepRegistryStub.sol
+++ b/contracts/test/KeepRegistryStub.sol
@@ -5,9 +5,9 @@ pragma solidity 0.8.9;
 import "./IKeepRegistry.sol";
 
 contract KeepRegistryStub is IKeepRegistry {
-    event OperatorContractApproved(address operatorContract);
-
     address public registryKeeper;
+
+    event OperatorContractApproved(address operatorContract);
 
     constructor() public {
         registryKeeper = msg.sender;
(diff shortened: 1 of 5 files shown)


deploy/08_configure_keep_registry.ts
```diff
@@ -0,0 +1,28 @@
...
+  const { getNamedAccounts, deployments } = hre
+  const { deployer } = await getNamedAccounts()
+  const { execute, log } = deployments
+
+  const TokenStaking = await deployments.get("TokenStaking")
+
+  await execute(
+    "KeepRegistry",
+    // TODO: We should execute this transaction from Keep deployer account. The
+    // Keep deployer account and T deployer account are 2 different accounts.
```
r-czajkowski, line 14: @michalinacienciala @nkuba I need help here.

r-czajkowski: ae83c15

contracts/test/KeepRegistryStub.sol
```diff
@@ -0,0 +1,19 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+pragma solidity 0.8.9;
+
+import "./IKeepRegistry.sol";
+
+contract KeepRegistryStub is IKeepRegistry {
+    event OperatorContractApproved(address operatorContract);
+
+    address public registryKeeper;
```
nkuba, line 10: Do we need this role in the stub at all? We're not checking the sender in `approveOperatorContract`.

deploy/08_configure_keep_registry.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  log(
+    `Approved T TokenStaking operator contract [${TokenStaking.address}] in Keep registry`
+  )
+}
+
+export default func
+
+func.tags = ["ConfigureKeepRegistry"]
+func.dependencies = ["TokenStaking", "KeepRegistry"]
```
nkuba, line 26: I guess we won't be able to run `approveOperatorContract` from the script for mainnet as it will have to be executed by the governance. Let's add skip for script then:

```ts
func.skip = async function (hre: HardhatRuntimeEnvironment): Promise<boolean> {
  return hre.network.name === "mainnet"
}
```

nkuba (changes requested):

contracts/test/KeepRegistryStub.sol
r-czajkowski: The `IKeepRegistry` interface requires it.

deploy/08_configure_keep_registry.ts
r-czajkowski: [8e7c5f0](https://github.com/threshold-network/solidity-contracts/pull/78/commits/8e7c5f04fff554e87dd6485563c45ad32568422e)

nkuba (approved):

PR #80, opened 2022-02-01
Preparation for deployment of TokenStaking

cygnusv:
* Modifies deployment script for TokenStaking for using OpenZeppelin Upgrades on mainnet
* Adds additional deployment steps for mainnet, since we need to transfer TokenStaking ownership and upgradeability control to the Threshold Council
* Adds the resulting OpenZeppelin Upgradeability Manifest, which is necessary to verify layout storage correctness of further TokenStaking upgrade

### DIFF
--- a/deploy/07_deploy_token_staking.ts
+++ b/deploy/07_deploy_token_staking.ts
@@ -1,8 +1,11 @@
 import { HardhatRuntimeEnvironment } from "hardhat/types"
 import { DeployFunction } from "hardhat-deploy/types"
 
+import { ethers, upgrades } from "hardhat"
+
 const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
   const { getNamedAccounts, deployments } = hre
+  const { execute, log } = deployments
   const { deployer } = await getNamedAccounts()
 
   const T = await deployments.get("T")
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@threshold-network/solidity-contracts",
-  "version": "1.1.0-dev",
+  "version": "1.1.0-alpha",
```
pdyraga, line 3: I think we should bump it to `1.2.0-dev` because all automated deployment flows we have expects `-dev` suffix AFAIK. Also, I'd bump it separately from this PR so that we can tag `v1.1.0` with the right version of deployment scripts. So what I'd do:

- review/merge this PR without bumping up version in `package.json`,
- create `releases/mainnet/v1.1.0` release branch,
- update dependencies to the ones from mainnet and version in `package.json` to `1.1.0` there and tag it as `v1.1.0`,
- create a PR that will bump up version to `1.2.0-dev` here.

hardhat.config.ts
```diff
@@ -97,10 +105,10 @@ const config: HardhatUserConfig = {
   namedAccounts: {
     deployer: {
       default: 0, // take the first account as deployer
-      mainnet: "0x123694886DBf5Ac94DDA07135349534536D14cAf",
+      // mainnet: "0x123694886DBf5Ac94DDA07135349534536D14cAf",
```
pdyraga, line 108: Should it be `0xfffd7092685bdeebd121d1a0fea3c349114cce50` for the most recent deployment?

cygnusv: TBH, I'm not sure what's the point of this value. We had some contracts deployed with `0x123694886DBf5Ac94DDA07135349534536D14cAf` (T, VendingMachines) and some with `0xfffd7092685bdeebd121d1a0fea3c349114cce50` (TokenStaking, proxies, KeepStake). I'd just remove this line.

package.json
cygnusv: Can you clarify what do you mean by `update dependencies to the ones from mainnet`? As for the rest, I think it's a good plan. We can leave the published 1.1.0-alpha just as a way to unblock the dashboard team.

pdyraga:
Should we have a step to init the staking contract proxy?

package.json
pdyraga: We have a dependency to `keep-core` that is pointing to `-dev` version on `main` (and I see it's also pointing to `-dev` in alpha package). We need to update dependency on the release branch to make sure correct contracts are resolved during the deployment and to have the mainnet NPM package referring to mainnet keep-core NPM package. This is for v1.0.0: https://github.com/threshold-network/solidity-contracts/blob/v1.0.0/package.json#L59-L66

hardhat.config.ts
pdyraga: Yeah, or just comment out and leave `""` to make our lives easier for the future (next deployments).

package.json
```diff
@@ -26,7 +27,7 @@
     "test": "hardhat test",
     "test:system": "NODE_ENV=system-test hardhat test ./test/system/*.test.js",
     "postinstall": "./scripts/prepare-dependencies.sh $INIT_CWD",
-    "prepack": "tsc -p tsconfig.export.json && hardhat export-artifacts export/artifacts",
+    "prepack": "tsc --target es5 -p tsconfig.export.json && hardhat export-artifacts export/artifacts",
```
nkuba, line 30: Can we set the target in `tsconfig.export.json` instead of adding `--target es5` arg here?

package.json
```diff
@@ -10,6 +10,7 @@
     "deploy/",
     "export/",
     "scripts/",
+    ".openzeppelin/mainnet.json",
```
nkuba, line 13: Can we define it as a directory, to not be tied to the mainnet?

```suggestion
    ".openzeppelin/",
```

hardhat.config.ts
cygnusv: Sure, we can do that, but I'm not sure if this is really used, since the script uses the private key from the `CONTRACT_OWNER_ACCOUNT_PRIVATE_KEY` envvar. FWIW, I deployed the contracts with the line commented and I had no problem with that.

deploy/07_deploy_token_staking.ts
```diff
@@ -12,23 +15,65 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
...
-      T.address,
-      KeepTokenStaking.address,
-      NuCypherStakingEscrow.address,
-      VendingMachineKeep.address,
-      VendingMachineNuCypher.address,
-      KeepStake.address,
-    ],
-    log: true,
-  })
+  const tokenStakingConstructorArgs = [
```
nkuba, line 18: `hardhat-deploy` already supports proxy deployments for Open Zeppelin (https://github.com/wighawag/hardhat-deploy#deploying-and-upgrading-proxies). Can't we use it instead of `"@openzeppelin/hardhat-upgrades"` plugin?

This would mean that we just need to add
```ts
    proxy: hre.network.name === "hardhat" ? false : {
      proxyContract: 'OpenZeppelinTransparentProxy'
    },
```
to `deployment.deploy` function options argument.

cygnusv: `hardhat-deploy` merely deploys the proper contracts, but OpenZeppelin Upgrades goes further to make sure that the contract is upgrade safe and that the storage layout is compatible with existing state. This implies checking the storage arrangement of the previous implementation with respect to any proposed upgrade
(see https://github.com/OpenZeppelin/openzeppelin-upgrades#how-do-the-plugins-work). To do so, it keeps track of storage variables in a separate file in `.openzeppelin/mainnet.json`, which I include in this PR.

package.json
cygnusv: Sure, I'll change it now.

package.json
cygnusv: OpenZeppelin produces upgradeability manifest files for each network (see https://docs.openzeppelin.com/upgrades-plugins/1.x/network-files#network.json). For example, I have locally a `.openzeppelin/unknown-31337.json` file, generated when I run the upgrades, and I don't want this file to be packaged. So, for the moment, I think it's better to tie it just to mainnet, as we only care for upgradeability there.

hardhat.config.ts
pdyraga: Ahh, I see. Well, then I'd vote to 🔥 it and maybe just leave a comment explaining it's read from `CONTRACT_OWNER_ACCOUNT_PRIVATE_KEY`.

pdyraga (approved):

Issue #88, opened 2022-04-11
Add Authorization-related events to the IApplication interface


cygnusv:
Now that there are drafts for real application contracts for Threshold, I just realized we never standardized some application-specific stuff like events for authorization operations. As an example, see the parallel evolution of `AuthorizationIncreased` in the PRE and tBTC app contracts, respectively:

```js
event AuthorizationIncreased(address indexed stakingProvider, uint96 fromAmount, uint96 toAmount);
```
(https://github.com/nucypher/nucypher/pull/2834/files#diff-a0f8e2c70e7a79b504a33807c59291816a3e3ed820aa1d195f79e500da6c861fR51)

vs.

``` js
event AuthorizationIncreased(address indexed stakingProvider, address indexed operator, uint96 fromAmount, uint96 toAmount);
```
(https://github.com/keep-network/keep-core/blob/main/solidity/ecdsa/contracts/WalletRegistry.sol#L184)

This can be problematic from interoperability and integrations point of view, particularly for the Threshold dashboard (cc @Battenfield ). As a solution, we need to find what's common on both implementations (considering also potential future application contracts), and add these events to the `IApplication` interface.

cygnusv:
> tBTC (ECDSA) and Random Beacon has a concept of an operator address that is running the node and the operator is automatically updating sortition pool state, that's why we included indexed operator address in the event.

@pdyraga How can generalize this for all Threshold apps? I don't think it's fully established that the role of operator is general; even though both tBTC and PRE have an operator, it may be the case in the future that some new app has 0 or more than 1 operators. Also, from the point of view of what an authorization is and what it entails, the operator role doesn't seem relevant.

Since `stakingProvider` is already indexed, can you remove the indexed `operator` field from the event and use the `self.stakingProviderToOperator` and `self.operatorToStakingProvider` mappings that you have in `EcdsaAuthorization`? Another solution would be to define a second event, completely app-specific, that complements the `AuthorizationIncreased`.

cygnusv:
**Current divergences:**
* `AuthorizationIncreased`:
    * Associated to `function authorizationIncreased()`  
    * tBTC has an extra `indexed operator` field in the event (see discussion in previous comment)
* `AuthorizationDecreaseRequested`:
    * Associated to `function authorizationDecreaseRequested()`
    * tBTC has an extra `uint64 decreasingAt` field in the event.
[...]

PR #91, opened 2022-04-21
Added minimumAuthorization function to IApplication interface

pdyraga:
Refs #88 
Refs https://github.com/keep-network/keep-core/issues/2935

~~Depends on #87 (keeping as a draft until #87 is merged)~~

There is an implicit contract in `TokenStaking` that `IApplication` may
revert in case the minimum authorization requirement is not met. The
problem is that `IApplication` does not clearly define what is the minimum
authorization.

This change improves it. First, the contract about
`increaseAuthorization` reverting if the minimum requirement is not met
is now explicit. Second, `IApplication` exposes `minimumAuthorization`
information.

### DIFF
--- a/contracts/staking/IApplication.sol
+++ b/contracts/staking/IApplication.sol
@@ -24,7 +24,9 @@ pragma solidity 0.8.9;
 interface IApplication {
     /// @notice Used by T staking contract to inform the application that the
     ///         authorized amount for the given staking provider increased.
-    ///         The application may do any necessary housekeeping.
+    ///         The application may do any necessary housekeeping. The
+    ///         application may revert the transaction in case the authorization
+    ///         is below the minimum required.
     function authorizationIncreased(
         address stakingProvider,
         uint96 fromAmount,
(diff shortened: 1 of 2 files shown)


contracts/staking/IApplication.sol
```diff
@@ -24,7 +24,9 @@ pragma solidity 0.8.9;
 interface IApplication {
     /// @notice Used by T staking contract to inform the application that the
     ///         authorized amount for the given staking provider increased.
-    ///         The application may do any necessary housekeeping.
+    ///         The application may do any necessary housekeeping. The
+    ///         application may revert the transaction in case the authorization
```
cygnusv, line 28:
```suggestion
    ///         application must revert the transaction in case the authorization
```
Otherwise, it doesn't make much sense to define it as a "minimum authorization"

cygnusv (approved):

vzotova (approved):

contracts/staking/IApplication.sol
pdyraga: Makes sense! Updated in the last commit.

pdyraga: Code format is complaining but I'll fix it once #87 is merged when rebasing over `main`. Ignoring for now.

pdyraga:
#87 got merged so I am rebasing this PR over `main` and undrafting!

pdyraga:
System tests are failing because of a Hardhat problem described here https://github.com/threshold-network/solidity-contracts/pull/87#issuecomment-1105343451.

PR #92, opened 2022-04-25
Fix deployment of KeepRegistryStub for external projects

nkuba:
When a KeepRegistry deployed contract is not found for local network we
wan to deploy a stub. The script can also be used by external projects
to deploy threshold network contract dependencies. When we used
`keepDeployer` the external projects deployment was failing as it
required the account to be defined also in the other project's hardhat
configuration file. To fix that we use `deployer` named account to
deploy the stub, as it will for sure be defined in other projects.

To call the `approveOperatorContract` we use `keepRegistryKeeper` which
is defined by default as account `0` and it matches the `deployer`.

### DIFF
--- a/hardhat.config.ts
+++ b/hardhat.config.ts
@@ -113,8 +113,8 @@ const config: HardhatUserConfig = {
     thresholdCouncil: {
       mainnet: "0x9F6e831c8F8939DC0C830C6e492e7cEf4f9C2F5f",
     },
-    keepDeployer: {
-      default: 0,
+    keepRegistryKeeper: {
+      default: 0, // same as the deployer
       ropsten: "0x923C5Dbf353e99394A21Aa7B67F3327Ca111C67D",
     },
   },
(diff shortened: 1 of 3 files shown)


hardhat.config.ts
```diff
@@ -113,8 +113,8 @@ const config: HardhatUserConfig = {
     thresholdCouncil: {
       mainnet: "0x9F6e831c8F8939DC0C830C6e492e7cEf4f9C2F5f",
     },
-    keepDeployer: {
-      default: 0,
+    keepRegistryKeeper: {
```
pdyraga, line 116: Maybe `keepRegistryOwner` or `keepRegistryGuvnor`? Keeper of keep registry sounds weird but if you don't think the alternatives are any better, I can also merge as-is.

pdyraga (approved):

hardhat.config.ts
nkuba: I agree it doesn't sound good, but I wanted to match the role that is defined in `KeepRegistry` that is named `registryKeeper`.

PR #95, opened 2022-04-26
Added reward-related functions to IApplication interface

### LINKED ISSUE #89: Generalize rewards-related methods and events for `IApplication`
In line with #88, I noticed there's no reference to rewards on `IApplication`, which implied that current prospective application contracts, namely PRE and tBTC, have their own implementation and interfaces. For the sake of interoperability with other dApps (e.g., dashboards), we should unify methods and events related to rewards and add them to `IApplication`.

To clarify, I'm not advocating for unifying the rewards distribution logic (although both PRE and tBTC follow the Unipool approach, a third application could follow a different rewards distribution mechanism), but for finding a generalized abstract interface that can be useful for all kinds of applications on Threshold. In very high-level terms, the functionalities of a Threshold application that are related to rewards are:

1. The Threshold protocol grants/pushes rewards to an application
1. The application has an internal mechanism to account for and distribute rewards to participating stakers
1. Stakers can query the amount of rewards they've accrued so far
1. Stakers can withdraw rewards from the application

I think all of these (except for functionality 2) should be generalized and added to `IApplication`.

pdyraga:
Closes #89

The interface is minimal and does not enforce any specific requirements on how the rewards are actually accrued.

`function withdrawRewards(address stakingProvider) external` withdraws application rewards for the given staking provider. Rewards are withdrawn to the staking provider's beneficiary address set in the staking contract. Function emits `RewardsWithdrawn(address indexed stakingProvider, uint96 amount)` so that applications can easily track the total amount of rewards accrued so far.

`function availableRewards(address stakingProvider)` Returns the amount of application rewards available for withdrawal for the given staking provider.

### DIFF
--- a/contracts/staking/IApplication.sol
+++ b/contracts/staking/IApplication.sol
@@ -22,6 +22,15 @@ pragma solidity 0.8.9;
 ///         staking provider are eligible to slash the stake delegated to that
 ///         staking provider.
 interface IApplication {
+    /// @dev Event emitted by `withdrawRewards` function.
+    event RewardsWithdrawn(address indexed stakingProvider, uint96 amount);
+
+    /// @notice Withdraws application rewards for the given staking provider.
+    ///         Rewards are withdrawn to the staking provider's beneficiary
+    ///         address set in the staking contract.
+    /// @dev Emits `RewardsWithdrawn` event.
+    function withdrawRewards(address stakingProvider) external;
+
     /// @notice Used by T staking contract to inform the application that the
     ///         authorized amount for the given staking provider increased.
     ///         The application may do any necessary housekeeping. The
@@ -56,6 +65,13 @@ interface IApplication {
         uint96 fromAmount,
         uint96 toAmount
     ) external;
+
+    /// @notice Returns the amount of application rewards available for
+    ///         withdrawal for the given staking provider.
+    function availableRewards(address stakingProvider)
+        external
+        view
+        returns (uint96);
 
     /// @notice The minimum authorization amount required for the staking
     ///         provider so that they can participate in the application.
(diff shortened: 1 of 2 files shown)


cygnusv:
> Function emits `RewardsWithdrawn(address indexed stakingProvider, uint96 amount)` so that applications can easily track the total amount of rewards accrued so far.

Although this doesn't affect the PR, just to note that the `RewardsWithdrawn` event can only track rewards that were withdrawn, not accrued.

In fact, I just realized if it makes sense to make a differentiation about accrued rewards and withdrawable rewards, e.g., because some app has some specific logic that limits withdrawals. In other words, is it possible that there's a context where, for some application, `availableRewards() < accruedRewards()`, instead of `availableRewards() == accruedRewards()`? If so, is that something we'd want to generalize?

vzotova (approved):

dimpar:
I take that there's no need for a function in this interface that withdraws only a part of accrued rewards?

Also, I don't think `RewardsWithdrawn` event will be enforced in the implementation, something that the devs will need to be aware of and not duplicate.

pdyraga:
> I take that there's no need for a function in this interface that withdraws only a part of accrued rewards?

I would not impose such requirements on applications. We want the reward withdrawal to be as cheap as possible, hence the minimum interface. If some application wants to provide something extra, sure but outside of the "must-have" contract.

> Also, I don't think `RewardsWithdrawn` event will be enforced in the implementation, something that the devs will need to be aware of and not duplicate.

There is no way we can enforce it on the compilation level but there is a `@dev` section in `withdrawRewards` docs that says this is the expected behavior.

cygnusv (approved):

PR #97, opened 2022-04-27
Define dependency to @openzeppelin/contracts as ~4.5

nkuba:
With `^4.5` a version `4.6.0` was resolved during installation which is incompatible due to changes in the `@openzeppelin/contracts/governance/Governor` contract.
We want to stick with `4.5` with patches, hence `~`.


cygnusv (approved):
Thanks for noticing this!

pdyraga (approved):

PR #99, opened 2022-04-28
Staking `requestAuthorizationDecrease` function contract update

pdyraga:
The application may revert authorization decrease request, if there is
already one pending for the given application.

This is entirely up to the application if overwriting pending request at
the given moment is fine or not.

This is important for TBTC and Random Beacon because allowing to
"decrease authorization decrease request" at any moment may be used for
pool manipulation and becoming a free-rider.

See https://github.com/keep-network/keep-core/pull/2960
See https://github.com/keep-network/keep-core/pull/2961


vzotova:
I thought when second request is send Application will try to start over decreasing process, so it's not overwriting but resetting. And if we always reset why it could lead to manipulation? (I'm fine if app can fail on such tx)

manumonti (approved):
I'm fine with this contribution, but I agree with Vicky's question

cygnusv (approved):

pdyraga:
> I thought when second request is send Application will try to start over decreasing process, so it's not overwriting but resetting. And if we always reset why it could lead to manipulation? (I'm fine if app can fail on such tx)

When the staker requests authorization decrease, their weight in the sortition pool is decreased right away. We do it to ensure the staker is not going to be selected to new signing groups based on their previous weight since after the delay period we'll not be able to slash their entire previous stake.

For example, if I have 100 tokens staked now and I request authorization decrease to 20 tokens, I should only be selected to new groups based on 20 tokens weight with an assumption that the protocol can slash me for no more than 20 tokens if I do something wrong in the new group.

If we allow to discard existing authorization decrease request and start a new one, the following manipulation is possible:
1. I have 100 tokens staked and I earn rewards for 100 tokens staked.
2. The new signing group selection is about to happen, I request authorization decrease to 1 token.
3. The new signing group selection selects me to the group based on 1 token weight (minimal responsibility).
4. I send a new authorization decrease request to 99 tokens. I now earn rewards for 99 tokens weight.

To prevent this manipulation, beacon and ECDSA can reject the authorization decrease request if the amount we are decreasing to is higher than the previous one. They can also accept the new request - with the delay being reset - if the amount to which we are decreasing to is lower than the previous one.

pdyraga:
The red build of system tests has been fixed in #103 at commit https://github.com/threshold-network/solidity-contracts/pull/103/commits/7c05a34a7b12453bc6e46146048eb484ccc5c799. The tests are passing, it's just some unrelated error returned by the runner after the tests are done executing. We can merge #103 first, then merge `main` to this PR and the build should turn green.

vzotova (approved):
Now I see, looks good!

PR #102, opened 2022-05-03
Removed hardhat-local-networks-config plugin

pdyraga:
After an update in `deepmerge` library, `hardhat-local-networks-config`
stopped working and that lead to our system tests failing with a
mysterious `TypeError: config.chains.has is not a function` error.

Until the problem is not fixed in `hardhat-local-networks-config`, the
plugin is removed. It was here for convenience and all tests are passing
without it.

See https://github.com/NomicFoundation/hardhat/issues/2630

To test this change, try running system tests on `main`:
```
FORKING_URL=... yarn test:system --verbose
```

Then, try doing the same on this branch.


hardhat.config.ts
```diff
@@ -1,7 +1,6 @@
 import { HardhatUserConfig } from "hardhat/config"
 
 import "@keep-network/hardhat-helpers"
-import "@keep-network/hardhat-local-networks-config"
```
nkuba, line 4: There is also a piece of configuration in lines 83-84 that has to be removed:
```
// // Define local networks configuration file path to load networks from the file.
// localNetworksConfig: "./.hardhat/networks.ts",
```

nkuba (changes requested):

nkuba (approved):

PR #103, opened 2022-05-06
Let the staking provider to refresh KEEP stake owner

pdyraga:
T network staking contract supports existing KEEP stakes by allowing KEEP stakers to use their stakes in T network and weights them based on KEEP<>T token ratio.

KEEP stake owner is cached in T staking contract and used to restrict access to all functions only the owner should be able to execute. To cache KEEP staking contract in T staking contract, T staking contract first needs to resolve the owner. Resolving liquid KEEP stake owner is easy. Resolving token grant stake owner is complicated and not possible to do on-chain from a contract external to KEEP `TokenStaking` contract. Keep `TokenStaking` knows the grant ID but does not expose it externally.

This problem is solved by `KeepStake` contract. It contains operator-owner pairs that were snapshotted 2021-11-11. If KEEP staker does not want to go liquid and instead they use their grant for delegation, and their grant was created after the snapshot date of 2021-11-11, `KeepStake` contract owner needs to register a new owner-operator pair for them.

The problem though is when the owner staked *before* `KeepStake` owner registered a pair of them.

An example is staking provider `0xA5f6822ef1A7DF72628259f9D1dc17eb2BCb2385` that received tokens delegated by owner `0x78dc7f84a8da5c96c4c54dcf5bd6e9ccd4fe2bb5` at transaction `0x262a7b173e2572e4494b3470c4978903e4d9f95141a18ad79c7d9090f16c9874`.

Before that delegation was done, no owner-pair was registered in `KeepStake` for `0x78dc7f84a8da5c96c4c54dcf5bd6e9ccd4fe2bb5` and `KeepStake.resolveOwner(0xA5f6822ef1A7DF72628259f9D1dc17eb2BCb2385)` returns `0xF139B3137D5bF24848e4F8FbE1d6AB70dF90c5Fa` which is the master Keep `TokenGrantStake` contract. As a result, `0xA5f6822ef1A7DF72628259f9D1dc17eb2BCb2385` can not execute calls as an owner.

The solution for this kind of situation would be to register the pair in `KeepStake` contract and call `TokenStaking.refreshKeepOwner()` but this function has a rule authorizing only the old owner to call this function. The owner is the master `TokenGrantStake` contract though.

The fix proposed here is to allow `TokenStaking.refreshKeepOwner()` call from the staking provider as well. We could probably also allow the authorizer to call it, but the staking provider feels good enough and such situations should be rare.


pdyraga:
System tests are in fact passing if you look at the execution log. It looks like some configuration problem, yet one issue after the fix from #102.

vzotova (approved):

manumonti (approved):
LGTM! ✌️

cygnusv (approved):

pdyraga:
@vzotova @manumonti @cygnusv I updated hardhat, ethers, and hardhat plugin versions in https://github.com/threshold-network/solidity-contracts/pull/103/commits/7c05a34a7b12453bc6e46146048eb484ccc5c799 and now system tests job is passing. System tests were passing before as well but the job was exiting with some test-unrelated error.

cygnusv (approved):

PR #107, opened 2022-05-18
Add bondOperator function to PRE stub

michalsmiarowski:
This change is needed to test token-dashboard properly.

Changes:
- add `bondOperator` function and `OperatorBonded` event to
`SimplePREApplicationStub` contract. 
- remove setting `operator` and `operatorStartTimestamp` values in 
`confirmOperatorAddress` since we are doing it in `bondOperator` func.


r-czajkowski (approved):
👍

PR #108, opened 2022-06-15
Adding `development` network for local development

dimpar:
We also need to run Geth locally for development purposes. For this, we need to allow `*Stub` contracts to be deployed on the local chain, otherwise, it will throw an error if the local network is not `hardhat`. This PR adds `development` network which runs on localhost and Geth can be run locally.

Refs https://github.com/keep-network/keep-core/issues/3011

### DIFF
--- a/deploy/00_resolve_keep_registry.ts
+++ b/deploy/00_resolve_keep_registry.ts
@@ -11,13 +11,9 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
   if (KeepRegistry && helpers.address.isValid(KeepRegistry.address)) {
     log(`using existing KeepRegistry contract at ${KeepRegistry.address}`)
   } else if (
-    hre.network.name === "hardhat" &&
-    (hre.network.config as HardhatNetworkConfig).forking.enabled
-  ) {
-    throw new Error("deployed KeepRegistry contract not found")
-  } else if (
-    hre.network.name !== "hardhat" &&
-    hre.network.name !== "development"
+    (hre.network.name === "hardhat" &&
+      (hre.network.config as HardhatNetworkConfig).forking.enabled) ||
+    (hre.network.name !== "hardhat" && hre.network.name !== "development")
   ) {
     throw new Error("deployed KeepRegistry contract not found")
   } else {
--- a/deploy/00_resolve_keep_registry.ts
+++ b/deploy/00_resolve_keep_registry.ts
@@ -11,8 +11,13 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
   if (KeepRegistry && helpers.address.isValid(KeepRegistry.address)) {
     log(`using existing KeepRegistry contract at ${KeepRegistry.address}`)
   } else if (
-    !hre.network.tags.local ||
+    hre.network.name === "hardhat" &&
     (hre.network.config as HardhatNetworkConfig).forking.enabled
+  ) {
+    throw new Error("deployed KeepRegistry contract not found")
+  } else if (
+    hre.network.name !== "hardhat" &&
+    hre.network.name !== "development"
   ) {
     throw new Error("deployed KeepRegistry contract not found")
   } else {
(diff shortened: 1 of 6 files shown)


deploy/00_resolve_keep_registry.ts
```diff
@@ -11,10 +11,15 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
...
     log(`using existing KeepRegistry contract at ${KeepRegistry.address}`)
   } else if (
-    hre.network.name !== "hardhat" ||
+    hre.network.name === "hardhat" &&
     (hre.network.config as HardhatNetworkConfig).forking.enabled
   ) {
     throw new Error("deployed KeepRegistry contract not found")
+  } else if (
+    hre.network.name !== "hardhat" &&
+    hre.network.name !== "development"
```
nkuba, line 20: `hardhat` and `development` networks are tagged `local` in the `hardhat.config.ts`. We could use `hre.network.tags.local` here.
```suggestion
    !hre.network.local
```

deploy/00_resolve_keep_registry.ts
```diff
@@ -11,10 +11,15 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
...
+  ) {
+    throw new Error("deployed KeepRegistry contract not found")
```
nkuba, line 22: We should try to combine these two `else if` statements.

deploy/00_resolve_keep_token.ts
```diff
@@ -16,8 +16,13 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
     // Save deployment artifact of external contract to include it in the package.
     await deployments.save("KeepToken", KeepToken)
   } else if (
-    hre.network.name !== "hardhat" ||
-    (hre.network.config as HardhatNetworkConfig).forking?.enabled
+    hre.network.name === "hardhat" &&
```
nkuba, line 19: Same comments as in `deploy/00_resolve_keep_registry.ts`.

deploy/00_resolve_keep_token_staking.ts
```diff
@@ -11,10 +11,15 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
   if (KeepTokenStaking && helpers.address.isValid(KeepTokenStaking.address)) {
     log(`using existing KeepTokenStaking at ${KeepTokenStaking.address}`)
   } else if (
-    hre.network.name !== "hardhat" ||
+    hre.network.name === "hardhat" &&
```
nkuba, line 14: Same comments as in `deploy/00_resolve_keep_registry.ts`.

nkuba (changes requested):

deploy/00_resolve_keep_registry.ts
dimpar: pls see these comments https://github.com/keep-network/tbtc-v2/pull/365#discussion_r897137774 
We already use hardhat's configuration in CI which has a `local` tag under the hood, but at the same time, CI is not a local machine so writing `!hre.network.local` when it runs on CI also doesn't hold 100%.

pdyraga: I would prefer to look at the network name in this condition. "local" tag is ambiguous. We may need more conditions for internal testnet and public testnet.

deploy/00_resolve_keep_token.ts
dimpar: pls see above

deploy/00_resolve_keep_token_staking.ts
dimpar: pls see above

deploy/00_resolve_keep_registry.ts
nkuba: Maybe we could use optional chaining?
`(hre.network.config as HardhatNetworkConfig)?.forking?.enabled`

dimpar: Yea, I like it, I didn't know about this optional chaining. [7005ee1](https://github.com/threshold-network/solidity-contracts/pull/108/commits/7005ee1aed1192fc1c70fe94c2b2f59b4fb1231a)

nkuba (approved):

deploy/00_resolve_keep_registry.ts
nkuba: We will use `allowStubs` tag.

PR #109, opened 2022-06-17
Update `NuCypherTokenStakingMock` contract

r-czajkowski:
Ref: https://github.com/threshold-network/token-dashboard/pull/127

Expose the `stakerInfo` function that returns the `staker` info to test
the T dapp locally and on Ropsten network. In the original impl of this
contract the `StakerStruct` is stored in the [`stakerInfo` map not
`stakers`](https://github.com/nucypher/nucypher-contracts/blob/main/contracts/contracts/StakingEscrow.sol#L151). Renaming the `stakers` to `stakerInfo` in stub contract
requires a lot of changes in tests so here we expose the `stakeInfo`
function to match the stub contract ABI with the original one.


pdyraga (approved):

contracts/test/TokenStakingTestSet.sol
```diff
@@ -182,6 +182,14 @@ contract NuCypherTokenStakingMock is INuCypherStakingEscrow {
...
         return stakers[staker].value;
     }
+
+    function stakerInfo(address staker)
+        public
+        view
+        returns (StakerStruct memory)
+    {
+        return stakers[staker];
+    }
```
pdyraga, line 192: Actually, one q: given that `stakers` is public, can't we query the contract with `contract.stakers(address)`?

PR #114, opened 2022-07-06
Add support for deployment on Goerli testnet

michalinacienciala:
Görli became a recommended test network after Ropsten's deprecation
notice (https://blog.ethereum.org/2022/06/21/testnet-deprecation/).
We're modifying GitHub Actions workflow for deploying contracts to
support the deployment on Görli. We're also leaving the possibility of
deployment on Ropsten (this will be removed once we have the Görli
deployment battle-tested and Ropsten gets shut down).
We also change the way we handle `keep-core` dependency - in V2 of our
system we'll no longer planning to publish new `keep-core` packages -
hence we hardcode used version of the package for given testnet.

TODO:

- [x] Configure `GOERLI_ETH_HOSTNAME_HTTP `, `GOERLI_ETH_CONTRACT_OWNER_PRIVATE_KEY`, `GOERLI_KEEP_ETH_CONTRACT_OWNER_PRIVATE_KEY`, `TENDERLY_TOKEN` GitHub secrets (@nkuba - could you do that? It would be also good to make similar ROPSTEN/GOERLI distinction in the secrets in the `keep-network` org)
- [x] Configure `CI_GITHUB_TOKEN` GitHub secret (@nkuba)
- [x] Verify the workflow once new V2-compatible config in `keep-core/ci` is ready
- [x] Remove testing configuration from the workflow

Refs:
https://github.com/keep-network/keep-core/issues/3012
https://github.com/keep-network/keep-core/pull/3081
https://github.com/keep-network/tbtc-v2/pull/382

### DIFF
--- a/.github/workflows/contracts.yaml
+++ b/.github/workflows/contracts.yaml
@@ -44,20 +44,19 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      # TODO: uncomment below block of code before merge to main
-      # - uses: actions/setup-node@v2
-      #   with:
-      #     node-version: "14.x"
-      #     cache: "yarn"
+      - uses: actions/setup-node@v2
+        with:
+          node-version: "14.x"
+          cache: "yarn"
 
-      # - name: Install dependencies
-      #   run: yarn install
+      - name: Install dependencies
+        run: yarn install
 
-      # - name: Build contracts
-      #   run: yarn build
+      - name: Build contracts
+        run: yarn build
 
-      # - name: Run tests
-      #   run: yarn test
+      - name: Run tests
+        run: yarn test
 
   contracts-system-tests:
     needs: contracts-detect-changes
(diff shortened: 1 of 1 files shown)


michalinacienciala:
> Everything except triggering next CI build works as expected. There's a problem with `CI_GITHUB_TOKEN` value - looks that the secret contains wrong credentials. Will need @pdyraga / @nkuba help with setting that correctly.

New value of `CI_GITHUB_TOKEN` has been set. Workflow has been verified again (in slightly shortened version, focusing on execution of `Notify CI...` step: https://github.com/threshold-network/solidity-contracts/runs/7341634837?check_suite_focus=true. Works as expected. PR ready for review.

.github/workflows/contracts.yaml
```diff
@@ -113,29 +113,19 @@ jobs:
...
-        id: upstream-builds-query
-        with:
-          upstream-builds: ${{ github.event.inputs.upstream_builds }}
-          query: keep-core-contracts-version = github.com/keep-network/keep-core/solidity-v1#version
-
       - name: Resolve latest contracts
-        run: |
-          yarn upgrade \
-            @keep-network/keep-core@${{ steps.upstream-builds-query.outputs.keep-core-contracts-version }}
+        run: yarn upgrade @keep-network/keep-core@1.8.1-goerli.0
```
nkuba, line 117: Could we get a package that is tagged with the environment name?

```suggestion
        run: yarn upgrade @keep-network/keep-core@${{ github.event.inputs.environment }}
```

michalinacienciala: Yes, we can. Resolving `keep-core` dependency using `goerli` tag did not work for me in another project, but here apparently it works fine. Changed in c9f33ca2306bf8298a97d9981540c9544f172393.

.github/workflows/contracts.yaml
```diff
@@ -99,7 +99,7 @@ jobs:
 
   contracts-deployment-testnet:
     needs: [contracts-build-and-test]
-    if: github.event_name == 'workflow_dispatch'
+    if: github.event_name == 'workflow_dispatch' && github.event.inputs.environment == 'goerli'
```
nkuba, line 102: Do we need to add `&& github.event.inputs.environment == 'goerli'`?
It's a manual workflow dispatch step, so maybe we could skip it?

michalinacienciala: [8c0c77d](https://github.com/threshold-network/solidity-contracts/pull/114/commits/8c0c77df74013532e4f049ea994806ae22abad18)

nkuba (approved):

PR #118, opened 2022-07-28
[Do not merge] dApp-friendly contract deployment

r-czajkowski:
Ref: https://github.com/threshold-network/token-dashboard/issues/136

We need a dApp-friendly version of deployment allowing dashboard developers to build UI components without having to run geth/ganache/hardhat locally and deploying contracts locally.

This PR aims at presenting changes to smart contracts deployed to Sepolia that are used by the dApp development team for day-to-day work. Please do not merge this PR to `main` branch. Smart contract with the changes presented here are pushed to NPM registry with a tag `dapp-development-sepolia`.

**Main changes:**
- sets the unstake period to 120s(2min) instead of 24h so we can test the unstake flow w/o waiting 24h,
- adds a script that sets the minimum stake amount to `39999999999999999999999` (only for the Sepolia network).


michalinacienciala:
Could you change this PR to draft and add a visible info in the description that those changes are not supposed to get merged to `main`?
Also, I was wondering if there's some way to additionally block the merging of this PR via GH settings. I mean something similar to protected branches functionality. From a quick googling I don't see such setting... Marking the PR as draft and explaining that the changes should not go to `main` will need to be enough then...

CODEOWNERS
```diff
@@ -0,0 +1 @@
+*.sol @pdyraga @nkuba @lukasz-zimnoch
```
pdyraga, line 1: Let's extract `CODEOWNERS` change to a separate PR and merge it to `main`. 

The `CODEOWNERS` should cover all existing contracts from `contracts/` directory

Let's add there all developers who worked on the Solidity code of the Threshold smart contracts: @vzotova @cygnusv @pdyraga 

This will not only protect dApp-dev-only changes from being merged but will also protect from any accidental changes to an already audited code.

deploy/09_set_minimum_stake.ts
```diff
@@ -0,0 +1,23 @@
...
+import { DeployFunction } from "hardhat-deploy/types"
+
+const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
+  const { getNamedAccounts, deployments } = hre
+  const { execute } = deployments
+  const { log } = deployments
+
+  const { deployer } = await getNamedAccounts()
+
+  const minStakeAmount = "39999999999999999999999"
```
pdyraga, line 11: Why do we want to set the minimum stake to 39999999999999999999999? With dApp team owning T token contract from `dapp-development-goerli` deployment, all devs should be able to mint any amount of T they want.

CODEOWNERS
r-czajkowski: https://github.com/threshold-network/solidity-contracts/pull/121

deploy/09_set_minimum_stake.ts
r-czajkowski: `39999999999999999999999` is the current min stake on mainnet that's why I picked this number but we could change to any number. We want to set the min stake to test the min stake validation in T dapp.

deploy/09_set_minimum_stake.ts
```diff
@@ -0,0 +1,23 @@
...
+  await execute("TokenStaking", { from: deployer }, "setMinimumStakeAmount")
```
michalinacienciala, line 12: When testing the deployment with the modified values, I got `Error: expected 1 arguments for method "setMinimumStakeAmount", got 0` error.
Full logs: https://github.com/threshold-network/solidity-contracts/runs/7782586514?check_suite_focus=true

r-czajkowski: c5bdf3d

michalinacienciala: After the fix the deployment worked: https://github.com/threshold-network/solidity-contracts/runs/7801852357?check_suite_focus=true.

mswilkison:
Closing this long-running demonstration PR, which is explicitly marked as not intended for merging into `main` and has had no commits since November 2023. The `dapp-development` branch is being retained so the deployment reference remains available. This closure does not merge the changes or delete the branch.

PR #119, opened 2022-07-28
Add job deploying contracts from `dapp-development` branch

michalinacienciala:
There are situations when team developing T Token Dashboard needs to
locally test some functionalities using modified contracts, for example
ones with shorter authorization decrease delay. We decided to create a
`dapp-development` branch in each of the upstream modules of
`threshold-network/token-dashboard` CI module, which would store the
code of these modified contracts. In this PR we create a
`contracts-dapp-development-deployment-testnet` job which deploys the
contracts, creates an NPM package (with `dappdev<environment>` suffix
an `dapp-development-<environment>` tag) and publishes it to the NPM
registry. At the end, the job also starts similar deployment for a
downstream module. The job gets triggered only as a result of
`workflow_dispath` event from a `dapp-development` branch. Currently
only `goerli` environment is supported. We don't run system and unit
tests for `dapp-development` branch, as the tests are not configured to
work with the modified contracts.
Generally, the goal of the changes is to have the full set of
dapp-development-friendly contracts deployed to the NPM registry, so
that the dApp developers could quickly use them by upgrading the
`token-dashboard` dependencies using `yarn upgrade
<package-name>@dapp-development-goerli`.
If the workflow gets dispatched from a different branch than
`dapp-development`, the deploy will behave as it used to, publishing
package with deployed unmodified contracts to the NPM registry under
`<environment>` tag.

TODO:

- [x] Modify the secret for the deployer account in the `contracts-dapp-development-deployment-testnet` job
- [x] Bump actions to latest versions
- [x] Test the workflow using combined code from #119 and #118 (`dapp-development` branch)

Refs:
https://github.com/threshold-network/token-dashboard/issues/136
https://github.com/keep-network/keep-core/pull/3121
https://github.com/keep-network/tbtc-v2/pull/392

### DIFF
--- a/.github/workflows/contracts.yaml
+++ b/.github/workflows/contracts.yaml
@@ -7,10 +7,24 @@ on:
     branches:
       - main
   pull_request:
+  # We intend to use `workflow dispatch` in two different situations/paths:
+  # 1. If a workflow will be manually dspatched from branch named
+  #    `dapp-development`, workflow will deploy the contracts on the selected
+  #    testnet and publish them to NPM registry with `dappdev<environment>`
+  #    suffix and `dapp-development-<environment>` tag. Such packages are meant
+  #    to be used locally by the team developing Threshold Token dApp and may
+  #    contain contracts that have different values from the ones used on
+  #    mainnet.
+  # 2. If a workflow will be manually dspatched from a branch which name is not
+  #    `dapp-development`, the workflow will deploy the contracts on the
+  #    selected testnet and publish them to NPM registry with `<environment>`
+  #    suffix and tag. Such packages will be used later to deploy public
+  #    Threshold Token dApp on a testnet, with contracts resembling those used
+  #    on mainnet.
   workflow_dispatch:
     inputs:
       environment:
-        description: "Environment for workflow execution"
+        description: "Environment (network) for workflow execution, e.g. `goerli`"
         required: false
         default: "dev"
       upstream_builds:
@@ -56,11 +70,14 @@ jobs:
         run: yarn build
 
       - name: Run tests
+        if: github.ref != 'refs/heads/dapp-development'
         run: yarn test
 
   contracts-system-tests:
     needs: contracts-detect-changes
-    if: needs.contracts-detect-changes.outputs.system-tests == 'true'
+    if: |
+      needs.contracts-detect-changes.outputs.system-tests == 'true'
+        && github.ref != 'refs/heads/dapp-development'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
@@ -99,7 +116,10 @@ jobs:
 
   contracts-deployment-testnet:
     needs: [contracts-build-and-test]
-    if: github.event_name == 'workflow_dispatch' && github.event.inputs.environment == 'goerli'
+    if: |
+      github.event_name == 'workflow_dispatch'
+        && github.event.inputs.environment == 'goerli'
+        && github.ref != 'refs/heads/dapp-development'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


.github/workflows/contracts.yaml
```diff
@@ -196,6 +216,67 @@ jobs:
...
+      - name: Notify CI about completion of the workflow
+        uses: keep-network/ci/actions/notify-workflow-completed@v2
+        env:
+          GITHUB_TOKEN: ${{ secrets.CI_GITHUB_TOKEN }}
+        with:
+          module: "github.com/threshold-network/solidity-contracts"
+          url: https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}
+          environment: ${{ github.event.inputs.environment }}
+          upstream_builds: ${{ github.event.inputs.upstream_builds }}
+          upstream_ref: dapp-development
```
michalinacienciala, line 277: I wondered whether to use explicit value (`dapp-development`) here, or `${{ github.event.inputs.upstream_ref }}`. Decided to use the former - it gives us less flexibility, but the general intention is that we always want to run the downstream modules on the same `dapp-development` branch. And by hardcoding it, we avoid the nasty consequences of the situation when somebody triggers the workflow from `dapp-development` branch, but with `upstream_ref` set to a different branch (which, if not for the harcoded value, would result in triggering of a downstream workflow from that different branch, resulting in publishing of `goerli`-tagged package that uses modified contracts in it's dependencies) .

michalinacienciala:
I dry-runned the workflow here: https://github.com/threshold-network/solidity-contracts/runs/7561186309?check_suite_focus=true. The workflow passed, but I noticed two things that seem strange to me:
1. Although in `contracts-dapp-development-deployment-testnet` job I don't have the step that prepares the Tenderly config, somehow the tenderly verification worked. I don't understand that. According to [hardhat-tenderly documentation](https://github.com/Tenderly/hardhat-tenderly)
> For this plugin to function you need to create a config.yaml file at `$HOME/.tenderly/config.yaml` or `%HOMEPATH%\.tenderly\config.yaml` and add an `access_key` field to it

Without the step that runs the script preparing that config, the file should not exist and Tenderly should not be able to authorize. But from the logs we see that it does. Either the plugin documentation is incomplete or I'm missing something...

2. In one of the runs I made when testing the new workflow structure, the contract deployment script did not mint the expected amount of `T` (`minted 0.0 T`). Here is the run: https://github.com/threshold-network/solidity-contracts/runs/7560768157?check_suite_focus=true.
I also had similar situation with `NU` (`minted 0.0 NU`) in another run:https://github.com/threshold-network/solidity-contracts/runs/7556698468?check_suite_focus=true.

michalinacienciala:
I tested the deployment again, this time on a branch being combination of changes from #119 and #118 (+ some small changes in workflows to allow for testing). Deployment has failed due to an issue in one of the modified deployment scripts, which I described in https://github.com/threshold-network/solidity-contracts/pull/118/files#r943208193.
The #119 can reviewed, but we will not be able to publish the package until the problem with failing deployment script gets fixed.

michalinacienciala:
> The https://github.com/threshold-network/solidity-contracts/pull/119 can reviewed, but we will not be able to publish the package until the problem with failing deployment script gets fixed.

The problem with the deployment script got fixed. I verified the #118 + #119 again, works as expected. Workflow run: https://github.com/threshold-network/solidity-contracts/runs/7801852357?check_suite_focus=true.

.github/workflows/contracts.yaml
```diff
@@ -196,20 +215,80 @@ jobs:
...
+          CHAIN_API_URL: ${{ secrets.GOERLI_ETH_HOSTNAME_HTTP }}
+          CONTRACT_OWNER_ACCOUNT_PRIVATE_KEY: ${{ secrets.DAPP_DEV_GOERLI_ETH_CONTRACT_OWNER_PRIVATE_KEY }}
+          KEEP_CONTRACT_OWNER_ACCOUNT_PRIVATE_KEY: ${{ secrets.GOERLI_KEEP_ETH_CONTRACT_OWNER_PRIVATE_KEY }}
+        run: yarn deploy --network ${{ github.event.inputs.environment }}
+
+      - name: Bump up package version
+        id: npm-version-bump
+        uses: keep-network/npm-version-bump@v2
+        with:
+          environment: dappdev${{ github.event.inputs.environment }}
```
nkuba, line 254: Could we try using underscores?

```suggestion
          environment: dapp_dev_${{ github.event.inputs.environment }}
```

nkuba: What do you think about using `*development*` instead of just `*dev*`? This would match the package tag more closely.
```suggestion
          environment: dapp_development_${{ github.event.inputs.environment }}
```

.github/workflows/contracts.yaml
```diff
@@ -196,20 +215,80 @@ jobs:
           yarn run hardhat --network ${{ github.event.inputs.environment }} \
             etherscan-verify --license GPL-3.0 --force-license
 
+  # This job is responsible for publishing packages from `dapp-development`
+  # branch, which are slightly modified to help with the process of testing some
+  # features on the Threshold Token dApp. The job starts only if workflow gets
+  # triggered by the `workflow_dispatch` event on the branch `dapp-development`.
+  contracts-dapp-development-deployment-testnet:
```
nkuba, line 222: I'm wondering whether common parts of `contracts-deployment-testnet` and `contracts-dapp-development-deployment-testnet` could be extracted to a reusable workflow or action. But this is something that could be handled in a follow-up PR.

.github/workflows/contracts.yaml
michalinacienciala: > Could we try using underscores?
I did try it some time ago, I was getting `invalid semver version: ${string}`

> What do you think about using *development* instead of just *dev*? This would match the package tag more closely.
I chose `dev` rather than `development` to be consistent with how we name the `development` tagged packages. We name them with `-dev.x` suffix, for exapmple `@keep-network/random-beacon@2.0.0-dev.64`.

.github/workflows/contracts.yaml
```diff
@@ -7,10 +7,24 @@ on:
...
+  #    suffix and tag. Such packages will be used later to deploy public
+  #    Threshold Token dApp on a testnet, with contracts resembling those used
+  #    on mainnet.
   workflow_dispatch:
     inputs:
       environment:
-        description: "Environment for workflow execution"
+        description: "Environment (network) for workflow execution, e.g. `goerli`"
         required: false
         default: "dev"
```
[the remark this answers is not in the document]
michalinacienciala: As a result of [conversation](https://github.com/threshold-network/token-dashboard/pull/157#discussion_r947926345) in one of PRs concerning other similar workflow I ended up removing the default value altogether. See 0e4e226b3a4d24fde8149047434c272bfb31b416.

.github/workflows/contracts.yaml
michalinacienciala: Definitely something we could do! There's more places in our workflows where using custom actions or reusable workflows could be useful. Let's not wait for these improvements to not postpone the merge. I will refactor the workflow in the future.

.github/workflows/contracts.yaml
nkuba: Please see https://github.com/keep-network/npm-version-bump/pull/3. Once it's merged we should be able to use `dapp-dev`.

michalinacienciala: [944d392](https://github.com/threshold-network/solidity-contracts/pull/119/commits/944d392512f9b2c3c78ee3885231b41cd17686a3)

nkuba (approved):

PR #126, opened 2022-09-22
Upgrade `hardhat-deploy` to latest version (`^0.11.15`)

michalinacienciala:
We want to upgrade `hardhat-deploy` dependency to latest version. Previously used version is not working correctly with a `@nomiclabs/hardhat-etherscan` plugin which we plan to use in the near future.
Apart from updating `package.lock` and `yarn.lock`, we had to change how we store the Etherscan key in Hardhat config file, because the syntax for that got changed in the newer versions of `hardhat-deploy`.

TODO:

- [x] fix failing unit tests

### DIFF
--- a/hardhat.config.ts
+++ b/hardhat.config.ts
@@ -88,8 +88,10 @@ const config: HardhatUserConfig = {
     username: "thesis",
     project: "thesis/threshold-network",
   },
-  etherscan: {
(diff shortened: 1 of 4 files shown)


michalinacienciala:
@nkuba, I reviewed your changes, LGTM. If you're happy with this PR as well, let's merge it.

Issue #150, opened 2023-07-26
Support deployment on Sepolia


michalinacienciala:
### Why?
The Görli testnet currently used by Threshold/Keep for development purposes is planned to become deprecated with the end of year 2023. The testnet that is planned to replace it is called [Holešky](https://github.com/eth-clients/holesky), however it's not yet available - it's planned it will become widely accessible on Oct 1, 2023 ([source](https://everstake.one/blog/new-ethereum-testnet-holesky-all-you-need-to-know-now)).
Switching our infrastructure to support new testnet is quite time consuming, so moving directly from Görli to Holešky may be quite risky, especially if there would be some delays in the date of Holešky genesis (not meeting the planned timelines is not a rare occurrence in the Ethereum space). As a solution, we decided to switch first to another testnet that is currently live - Sepolia. This testnet's EOL is planned for 2026, which gives us plenty of time to move to Holešky before Sepolia gets deprecated.

**Testnets Lifecycle:**
![image](https://github.com/threshold-network/solidity-contracts/assets/78352137/fca11ea0-ad73-43e1-af33-e1aed24a545b)
Read more in [Predictable Ethereum Testnet Lifecycle Proposal](https://ethereum-magicians.org/t/proposal-predictable-ethereum-testnet-lifecycle/11575).

### How?
Deploying contracts to a new testnet and adjusting CI accordingly is a work consisting of many tasks and requiring coordination of several team members. We shouldn't get rid of Görli support until we're sure that everything works on Sepolia smoothly (maybe even until Görli gets deprecated). We may consider adding Holešky support together with Sepolia in some places, for the future use (for example in Hardhat config).

### What & who?

- [x] **Update** list of supported networks in **`keep-common`** (https://github.com/keep-network/keep-common/pull/118)
- [ ] Adjust our **Hardhat/Truffle configuration** to support Sepolia (without removing configuration for Görli yet):
  - [x] `@keep-network/keep-core` - https://github.com/keep-network/keep-core/pull/3712
  - [x] `@keep-network/keep-ecdsa` - modified locally (repository deprecated)
  - [x] `@keep-network/tbtc` - modified locally (repository deprecated)
  - [x] `@threshold-network/solidity-contracts` - https://github.com/threshold-network/solidity-contracts/pull/151, https://github.com/threshold-network/solidity-contracts/pull/152, https://github.com/threshold-network/solidity-contracts/pull/154
[...]

PR #151, opened 2023-09-05
Support deployment on Sepolia

michalinacienciala:
The Görli testnet currently used by Threshold/Keep for development purposes is planned to become deprecated with the end of year 2023. The testnet that is planned to replace it is called
[Holešky](https://github.com/eth-clients/holesky), however it's not yet available - it's planned it will become widely accessible on Oct 1, 2023 ([source](https://everstake.one/blog/new-ethereum-testnet-holesky-all-you-need-to-know-now)). Switching our infrastructure to support new testnet is quite time consuming, so moving directly from Görli to Holešky may be quite risky, especially if there would be some delays in the date of Holešky genesis (not meeting the planned timelines is not a rare occurrence in the Ethereum space). As a solution, we decided to switch first to another testnet that is currently live - Sepolia. This testnet's EOL is planned for 2026, which gives us plenty of time to move to Holešky before Sepolia gets deprecated.

Refs:
https://github.com/threshold-network/solidity-contracts/issues/150
https://github.com/keep-network/ci/pull/48
https://github.com/keep-network/keep-core/pull/3706
https://github.com/keep-network/tbtc-v2/pull/691
https://github.com/threshold-network/token-dashboard/pull/605

TODO:

- [ ] Configure secrets:
```
TESTNET_ETH_CONTRACT_OWNER_PRIVATE_KEY
DAPP_DEV_TESTNET_ETH_CONTRACT_OWNER_PRIVATE_KEY
SEPOLIA_ETH_HOSTNAME_HTTP
```


.github/workflows/contracts.yml
```diff
@@ -153,9 +153,22 @@ jobs:
 
       - name: Deploy contracts
         env:
-          CHAIN_API_URL: ${{ secrets.GOERLI_ETH_HOSTNAME_HTTP }}
-          CONTRACT_OWNER_ACCOUNT_PRIVATE_KEY: ${{ secrets.GOERLI_ETH_CONTRACT_OWNER_PRIVATE_KEY }}
-          KEEP_CONTRACT_OWNER_ACCOUNT_PRIVATE_KEY: ${{ secrets.GOERLI_KEEP_ETH_CONTRACT_OWNER_PRIVATE_KEY }}
+          # Use fake ternary expressions to decide which credentials to use,
+          # depending on chosen environment. Note: if `GOERLI...` credentials
+          # are empty, the expressions will be evaluated to the `SEPOLIA...`
+          # ones.
```
lukasz-zimnoch, line 159: If `environment` is not set at all (it can be as this input is not required), we also fallback to `SEPOLIA_*` right? Is this a desired behavior? If so, can we mention about it here as well?

michalinacienciala: It makes sense to make `environment` a required input, I did it in 440c305e7f68d536b561a1fdc4091c67cd96789f. I also removed some of the ternary expressions (in e2c80e627aab48022a3fe66b1ab924ba654784d4), as we are reusing the same deployer address for Goerli and Sepolia.

michalinacienciala:
@lukasz-zimnoch, I've realized I was missing some changes in order to make a successful deployment and added a couple of commits (and verified locally that contracts now deploy on Sepolia). I wasn't 100% sure if the files in `.external/sepolia` can be just copied from the `keep-core` deployment or should I do any changes to them (e.g. remove `address` or `networks`) - I did no changes and the deployment worked - but if there's some reason I should remove some data, please let me know.
Also, if you have the right privileges (I still am missing them), could you set the environment variables listed in the TODO section of the PR summary?

lukasz-zimnoch:
> @lukasz-zimnoch, I've realized I was missing some changes in order to make a successful deployment and added a couple of commits (and verified locally that contracts now deploy on Sepolia). I wasn't 100% sure if the files in `.external/sepolia` can be just copied from the `keep-core` deployment or should I do any changes to them (e.g. remove `address` or `networks`) - I did no changes and the deployment worked - but if there's some reason I should remove some data, please let me know. Also, if you have the right privileges (I still am missing them), could you set the environment variables listed in the TODO section of the PR summary?

Normally, those files should be copied from the `keep-core` Sepolia deployment info. This project uses them as a reference to the legacy Keep contracts. If they are copied from https://www.npmjs.com/package/@keep-network/keep-core/v/1.8.1-sepolia.0 I think we are good. Just make sure they have the `address` field. I see that is the case for Goerli artifacts and those for Sepolia have `networks.11155111.address` which is a bit different. But if that works then 🤷 - maybe some changes in our Hardhat configs were done in the meantime.

michalinacienciala:
> If they are copied from https://www.npmjs.com/package/@keep-network/keep-core/v/1.8.1-sepolia.0 I think we are good.

Yes, the are.

>  I see that is the case for Goerli artifacts and those for Sepolia have networks.11155111.address which is a bit different. But if that works then 🤷 - maybe some changes in our Hardhat configs were done in the meantime.
EDIT: You can make sure the external contract addresses are resolved properly by console logging KeepTokenStaking.address in the deploy/07_deploy_token_staking.ts file during deployment

Confirmed, the right TokenStaking address was used (`0x91Fe7128f74dBd4F031ea3D90FC5Ea4DCfD81818`).

Package deployed and published successfully: https://www.npmjs.com/package/@threshold-network/solidity-contracts/v/1.3.0-sepolia.0.
I think we can merge once we have the secrets set up.

lukasz-zimnoch (approved):

PR #153, opened 2023-09-19
First step to remove legacy

vzotova:
Removes most of legacy support: staking Keep/Nu, top-up Keep/Nu, increasing authorization to more than T stake.
Also adds method to deauthorize+unstake legacy stakes in one tx using involuntary decrease call. This method is permissionless.
Description of how this all would work can be found here #141. Additional to that - `forceUnstakeLegacy` method.
Second step will be removing `forceUnstakeLegacy` and all other usage of legacy stakes in `TokenStaking` after unstaking all legacy tokens.
P.S. `stakedNu` used in NuCypher `StakingEscrow`, for compatibility and will be removed at step 2 (together with link to Nu vending machine)

PR #110, opened 2022-06-25
Allow anyone to top-up any staking provider with liquid T

vzotova:

PR #155, opened 2023-09-29
Adds `autoIncrease` flag to increase authorizations in `topUp`

vzotova:
Based on #153 
With flag set to true each `topUp` will increase authorization amount for all authorized app

### DIFF
--- a/contracts/staking/IStaking.sol
+++ b/contracts/staking/IStaking.sol
@@ -158,9 +158,15 @@ interface IStaking {
     //
 
     /// @notice Increases the amount of the stake for the given staking provider.
+    ///         If `autoIncrease` flag is true then the amount will be added for
+    ///         all authorized applications.
     /// @dev The sender of this transaction needs to have the amount approved to
     ///      transfer to the staking contract.
     function topUp(address stakingProvider, uint96 amount) external;
+
+    /// @notice Toggle auto authorization increase flag. If true then all amount
+    ///         in top-up will be added to already authorized applications.
+    function toggleAutoAuthorizationIncrease(address stakingProvider) external;
 
     //
     //
@@ -277,6 +283,12 @@ interface IStaking {
         external
         view
         returns (uint256);
+
+    /// @notice Returns auto-increase flag.
+    function getAutoIncreaseFlag(address stakingProvider)
+        external
+        view
+        returns (bool);
 
     /// @notice Returns staked amount of NU for the specified staking provider.
     function stakedNu(address stakingProvider) external view returns (uint256);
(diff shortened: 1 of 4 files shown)


cygnusv (approved):
LGTM! Just one comment: Why using the "toggle" approach instead of a setter? It's slightly less ergonomic, in my opinion, but interested to see what was your line of thinking.

contracts/staking/IStaking.sol
```diff
@@ -177,19 +158,15 @@ interface IStaking {
...
-    ///         the owner or the staking provider.
-    function topUpKeep(address stakingProvider) external;
-
-    /// @notice Propagates information about stake top-up from the legacy NU
-    ///         staking contract to T staking contract. Can be called only by
-    ///         the owner or the staking provider.
-    function topUpNu(address stakingProvider) external;
+    /// @notice Toggle auto authorization increase flag. If true then all amount
+    ///         in top-up will be added to already authorized applications.
+    function toggleAutoAuthorizationIncrease(address stakingProvider) external;
```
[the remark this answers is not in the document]
vzotova: I'm not sure about that, I want names explicit as possible

vzotova:
> LGTM! Just one comment: Why using the "toggle" approach instead of a setter? It's slightly less ergonomic, in my opinion, but interested to see what was your line of thinking.

not strong opinion: from code perspective - if you call this tx you want to change parameter/toggle, there is no other options that you can change except to opposite value. In the same case setter should fail when parameter is equal.

cygnusv (approved):

contracts/staking/TokenStaking.sol
```diff
@@ -590,6 +597,54 @@ contract TokenStaking is Initializable, IStaking, Checkpoints {
         emit ToppedUp(stakingProvider, amount);
         increaseStakeCheckpoint(stakingProvider, amount);
         token.safeTransferFrom(msg.sender, address(this), amount);
+
+        if (!stakingProviderStruct.autoIncrease) {
+            return;
+        }
+
+        // increase authorization for all authorized app
+        for (
```
lukasz-zimnoch, line 606: The `topUp` function can be called by anyone so, if `autoIncrease` is `true`, anyone can increase staking provider's authorization. Of course, the caller has to transfer an amount, but still, wondering if we aren't opening a potential attack vector here. Shouldn't we do the increase just for callers being provider's authorizers/owners? Curious about your thoughts!

cygnusv: Yes, this was intentionally added in https://github.com/threshold-network/solidity-contracts/pull/110.
The rationale was to allow other contracts like Merkle distribution or application contracts to automatically top up stakes.

lukasz-zimnoch: Ok, thanks for clarification!

lukasz-zimnoch (dismissed):

theref (approved):

cygnusv (approved):

PR #157, opened 2023-11-10
Adding interface for applications that use operator role and delay in…

### LINKED ISSUE #156: Extend IApplication to generalize tBTC, RandomBeacon and TACo apps
In addition to the discussion on #88, while developing the TACo app contract, and particularly when integrating it with the dashboard, we've detected several methods that should be generalized in `IApplication` due to common workflows. This is specially relevant in contexts that consume the 3 contracts (i.e. the dashboard).

Also, consider if it's necessary the creation of an additional interface for Threshold apps that assume the existence of an operator role, something like `IApplicationWithOperator`, so we can abstract here methods like `registerOperator()`.

vzotova:
… authorization decrease.

Fixes #156

Main suggestion here is to use both interfaces for all apps: TBCv2, RandomBeacon, TACo.
Names for method inherit those from `RandomBeacon` and `TBTCv2`, except one additional method to register operator using additional role (please leave comment: remove or keep it).
That unification will help with tracking changes between TBTC and TACo contracts, simplify dashboard interaction with contracts and also help with contract pool developing in the future.

### DIFF
--- a/contracts/staking/IApplicationWithDecreaseDelay.sol
+++ b/contracts/staking/IApplicationWithDecreaseDelay.sol
@@ -1,0 +1,68 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+// ██████████████     ▐████▌     ██████████████
+// ██████████████     ▐████▌     ██████████████
+//               ▐████▌    ▐████▌
+//               ▐████▌    ▐████▌
+// ██████████████     ▐████▌     ██████████████
+// ██████████████     ▐████▌     ██████████████
+//               ▐████▌    ▐████▌
+//               ▐████▌    ▐████▌
+//               ▐████▌    ▐████▌
+//               ▐████▌    ▐████▌
+//               ▐████▌    ▐████▌
+//               ▐████▌    ▐████▌
+
+pragma solidity ^0.8.9;
+
+import "./IApplication.sol";
+
+/// @title  Interface for Threshold Network applications with delay after decrease request
+interface IApplicationWithDecreaseDelay is IApplication {
+    /// @notice Returns authorization-related parameters of the application.
+    /// @dev The minimum authorization is also returned by `minimumAuthorization()`
+    ///      function, as a requirement of `IApplication` interface.
+    /// @return _minimumAuthorization The minimum authorization amount required
+    ///         so that operator can participate in the application.
+    /// @return authorizationDecreaseDelay Delay in seconds that needs to pass
+    ///         between the time authorization decrease is requested and the
+    ///         time that request gets approved. Protects against free-riders
+    ///         earning rewards and not being active in the network.
+    /// @return authorizationDecreaseChangePeriod Authorization decrease change
+    ///        period in seconds. It is the time, before authorization decrease
+    ///        delay end, during which the pending authorization decrease
+    ///        request can be overwritten.
+    ///        If set to 0, pending authorization decrease request can not be
+    ///        overwritten until the entire `authorizationDecreaseDelay` ends.
+    ///        If set to value equal `authorizationDecreaseDelay`, request can
+    ///        always be overwritten.
+    function authorizationParameters()
+        external
+        view
+        returns (
+            uint96 _minimumAuthorization,
+            uint64 authorizationDecreaseDelay,
+            uint64 authorizationDecreaseChangePeriod
+        );
+
+    /// @notice Returns the amount of stake that is pending authorization
+    ///         decrease for the given staking provider. If no authorization
+    ///         decrease has been requested, returns zero.
+    function pendingAuthorizationDecrease(address _stakingProvider)
+        external
+        view
+        returns (uint96);
+
+    /// @notice Returns the remaining time in seconds that needs to pass before
+    ///         the requested authorization decrease can be approved.
+    function remainingAuthorizationDecreaseDelay(address stakingProvider)
+        external
+        view
+        returns (uint64);
+
+    /// @notice Approves the previously registered authorization decrease
+    ///         request. Reverts if authorization decrease delay has not passed
+    ///         yet or if the authorization decrease was not requested for the
+    ///         given staking provider.
+    function approveAuthorizationDecrease(address stakingProvider) external;
+}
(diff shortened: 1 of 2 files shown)


theref (approved):

contracts/staking/IApplicationWithOperator.sol
```diff
@@ -0,0 +1,42 @@
...
+/// @title  Interface for Threshold Network applications with operator role
+interface IApplicationWithOperator is IApplication {
+    /// @notice Returns operator registered for the given staking provider.
+    function stakingProviderToOperator(address stakingProvider)
+        external
+        view
+        returns (address);
+
+    /// @notice Used by staking provider to set operator address that will
+    ///         operate a node. The operator addressmust be unique.
```
cygnusv, line 29:
```suggestion
    ///         operate a node. The operator address must be unique.
```

contracts/staking/IApplicationWithOperator.sol
```diff
@@ -0,0 +1,42 @@
...
+    ///         Reverts if the operator is already set for the staking provider
+    ///         or if the operator address is already in use.
+    /// @dev    Depending on application the given staking provider can set operator
+    ///         address only one or multiple times. Besides that application can decide
```
cygnusv, line 33:
```suggestion
    ///         address only once or multiple times. Besides that, application can decide
```

contracts/staking/IApplicationWithOperator.sol
```diff
@@ -0,0 +1,42 @@
...
+    ///         if function reverts if there is a pending authorization decrease for
+    ///         the staking provider.
+    function registerOperator(address operator) external;
+
+    // TODO consider that?
+    // /// @notice Used by additional role (owner for example) to set operator address that will
+    // ///         operate a node for the specified staking provider.
+    // function registerOperator(address stakingProvider, address operator) external;
```
cygnusv, line 41: What's your opinion on this @vzotova @lukasz-zimnoch ? My suggestion would be to not include it in the interface, so it can be optionally implemented by applications that allow owners to set operators.

contracts/staking/IApplicationWithDecreaseDelay.sol
```diff
@@ -0,0 +1,68 @@
...
+    ///        request can be overwritten.
+    ///        If set to 0, pending authorization decrease request can not be
+    ///        overwritten until the entire `authorizationDecreaseDelay` ends.
+    ///        If set to value equal `authorizationDecreaseDelay`, request can
+    ///        always be overwritten.
+    function authorizationParameters()
+        external
+        view
+        returns (
+            uint96 _minimumAuthorization,
```
cygnusv, line 43: Why the `_` here? Same for `_stakingProvider` in L51.

contracts/staking/IApplicationWithDecreaseDelay.sol
```diff
@@ -0,0 +1,68 @@
...
+/// @title  Interface for Threshold Network applications with delay after decrease request
+interface IApplicationWithDecreaseDelay is IApplication {
+    /// @notice Returns authorization-related parameters of the application.
+    /// @dev The minimum authorization is also returned by `minimumAuthorization()`
+    ///      function, as a requirement of `IApplication` interface.
+    /// @return _minimumAuthorization The minimum authorization amount required
+    ///         so that operator can participate in the application.
+    /// @return authorizationDecreaseDelay Delay in seconds that needs to pass
+    ///         between the time authorization decrease is requested and the
+    ///         time that request gets approved. Protects against free-riders
```
cygnusv, line 29: Maybe free-riders is a bit strong, considering legitimate scenarions like normal deauthorizations.
```suggestion
    ///         time that request gets approved. Protects against participants
```

contracts/staking/IApplicationWithDecreaseDelay.sol
```diff
@@ -0,0 +1,68 @@
...
+    ///         earning rewards and not being active in the network.
+    /// @return authorizationDecreaseChangePeriod Authorization decrease change
+    ///        period in seconds. It is the time, before authorization decrease
+    ///        delay end, during which the pending authorization decrease
```
cygnusv, line 33:
```suggestion
    ///        period in seconds. It is the time window, before authorization decrease
    ///        delay ends, during which the pending authorization decrease
```

contracts/staking/IApplicationWithDecreaseDelay.sol
vzotova: otherwise it shadows `function minimumAuthorization`

contracts/staking/IApplicationWithOperator.sol
vzotova: I'm good with both options

lukasz-zimnoch: > What's your opinion on this @vzotova @lukasz-zimnoch ? My suggestion would be to not include it in the interface, so it can be optionally implemented by applications that allow owners to set operators.

Agreed! That feels optional and should not live on the interface level.

theref: Then we need to remove it here as well? https://github.com/threshold-network/token-dashboard/blob/bbdc89047f2f32905cf35ccbb07ac414dc9b0ab2/src/threshold-ts/applications/index.ts#L191

We should probably be using the `solidity-contracts` repo in the dashboard rather than having a bunch of interfaces hard coded there

vzotova: ok, removed
@theref there is "regular" `registerOperator` in your link and this method still in interface

theref: oh yes sorry, hadn't noticed the difference in signature.

ok, well the dashboard is using an interface to specify how application contracts should behave. If we don't include the version that allows applications to set Operators then it's not going to be available on the dashboard (unless we completely change the dashboard architecture)

manumonti (approved):
LGTM! 👌

theref (approved):

cygnusv (approved):

PR #159, opened 2023-11-27
Second step to remove legacy

vzotova:
Now when all legacy token unstaked we don't need any support of it. 
All methods related to legacy were removed including new one to force unstake.


contracts/staking/IStaking.sol
```diff
@@ -176,39 +170,11 @@ interface IStaking {
...
-    ///         from NU staking contract while still being able to operate in
-    ///         T network and earning rewards based on the native T staked.
-    ///         Can be called only by the stake owner or the staking provider.
-    function unstakeNu(address stakingProvider) external;
-
-    /// @notice Sets cached legacy stake amount to 0, sets the liquid T stake
-    ///         amount to 0 and withdraws all liquid T from the stake to the
-    ///         owner. Reverts if there is at least one non-zero authorization.
+    ///         authorization higher than the remaining liquid T stake or
+    ///         if the unstake amount is higher than the liquid T stake amount.
```
cygnusv, line 174: I think we can change "`liquid T stake`" (which doesn't make much sense now) into just "`T stake`" everywhere

contracts/staking/IStaking.sol
```diff
@@ -265,17 +231,8 @@ interface IStaking {
...
-    function stakes(address stakingProvider)
-        external
-        view
-        returns (
-            uint96 tStake,
-            uint96 keepInTStake,
-            uint96 nuInTStake
-        );
+    /// @notice Returns staked amount of T for the specified staking provider.
+    function tStake(address stakingProvider) external view returns (uint96);
```
cygnusv, line 235: What about `stakeAmount`, or even just `stake`?

contracts/staking/TokenStaking.sol
```diff
@@ -28,12 +28,8 @@ import "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
 import "@openzeppelin/contracts-upgradeable/utils/AddressUpgradeable.sol";
 
 /// @notice TokenStaking is the main staking contract of the Threshold Network.
-///         Apart from the basic usage of enabling T stakes, it also acts as a
-///         sort of "meta-staking" contract, accepting existing legacy NU/KEEP
-///         stakes. Additionally, it serves as application manager for the apps
-///         that run on the Threshold Network. Note that legacy NU/KEEP staking
-///         contracts see TokenStaking as an application (e.g., slashing is
-///         requested by TokenStaking and performed by the legacy contracts).
+///         Additionally, it serves as application manager for the apps
```
cygnusv, line 31:
```suggestion
///         It serves as application manager for the apps
```

contracts/staking/TokenStaking.sol
```diff
@@ -108,7 +101,6 @@ contract TokenStaking is Initializable, IStaking, Checkpoints {
     uint256 public slashingQueueIndex;
 
     event Staked(
-        StakeType indexed stakeType,
```
cygnusv, line 111: I'm afraid we can't drop this. There's already a bunch of external apps (e.g., Dune queries) that consume events in their current form.

cygnusv (changes requested):

contracts/staking/TokenStaking.sol
```diff
@@ -51,9 +47,9 @@ contract TokenStaking is Initializable, IStaking, Checkpoints {
     }
 
     struct StakingProviderInfo {
-        uint96 nuInTStake;
+        uint96 legacyNuInTStake;
         address owner;
-        uint96 keepInTStake;
+        uint96 legacyKeepInTStake;
```
cygnusv, line 52: Can we leave the old names? I ran the OZ upgrade tool on the new contract (see below), and so far the only thing it doesn't like is this name change. For variable names, there's a way to override this error, but for the moment not for struct members ( see https://docs.openzeppelin.com/upgrades-plugins/1.x/faq#how-to-rename). For the moment, instead of changing the name we can leave a comment.

OZ output:
```
Error: ERROR processing /Users/david/dev/t/solidity-contracts/deploy/60_validate_upgrade_token_staking.ts:
Error: New storage layout is incompatible

contracts/staking/TokenStaking.sol:120: Upgraded `stakingProviders` to an incompatible type
  - In mapping(address => struct TokenStaking.StakingProviderInfo)
    - Bad upgrade to struct TokenStaking.StakingProviderInfo
  - In struct TokenStaking.StakingProviderInfo
    - Renamed `nuInTStake` to `legacyNuInTStake`
    - Renamed `keepInTStake` to `legacyKeepInTStake`
```

Side-note: I need to add this validation tool to CI, probably I'm doing it as part of #140. I also need to check why the tool doesn't complain about the change in the `Staked` event signature.

cygnusv (changes requested):

contracts/staking/IStaking.sol
vzotova: `stake` is used, I'm fine with `stakeAmount`

vzotova:
> Do we still need the `@keep-network/keep-core` dependency in [`package.json`](https://github.com/threshold-network/solidity-contracts/blob/remove-legacy-step-2/package.json#L61)?

Didn't find any usage so removed, could you check please that nothing needed was not removed?

contracts/staking/IStaking.sol
vzotova: ✔️

contracts/staking/TokenStaking.sol
vzotova: ✔️

contracts/staking/TokenStaking.sol
vzotova: Got it, added back ✔️

contracts/staking/TokenStaking.sol
vzotova: Changing of event can't break storage but renaming theoretically can (if we would use those variables) somewhere. 
Renamed back ✔️

.github/workflows/contracts.yml
```diff
@@ -197,7 +194,6 @@ jobs:
       # `etherscan-verify` plugins tries to verify them, which is not desired.
       - name: Prepare for verification on Etherscan
         run: |
-          rm -rf ./node_modules/@keep-network/keep-core
           rm -rf ./external/npm
```
michalinacienciala, line 197: We can remove this whole step (the `rm -rf ./external/npm` become obsolete some time ago).

.github/workflows/npm.yml
```diff
@@ -27,11 +27,6 @@ jobs:
           registry-url: "https://registry.npmjs.org"
           cache: "yarn"
 
-      - name: Resolve latest contracts
-        run: |
-          yarn upgrade --exact \
-            @keep-network/keep-core
-
```
michalinacienciala, line 34: If we're removing `yarn upgrade ...` we need to add `yarn install --frozen-lockfile` in order to install dependencies.

michalinacienciala (changes requested):
> > Do we still need the @keep-network/keep-core dependency in [package.json](https://github.com/threshold-network/solidity-contracts/blob/remove-legacy-step-2/package.json#L61)?

> Didn't find any usage so removed, @michalinacienciala could you check please that nothing needed was not removed?

I left two comments. I think we can also do a cleanup of Keep contracts in the `./external` folder and remove these lines:
https://github.com/threshold-network/solidity-contracts/blob/5fb485ec28ae6abc8544b2ad0ede1fc83b663537/hardhat.config.ts#L94-L95.

.github/workflows/contracts.yml
vzotova: ✔️

.github/workflows/npm.yml
vzotova: ✔️

hardhat.config.ts
```diff
@@ -91,8 +91,6 @@ const config: HardhatUserConfig = {
       // For hardhat environment we can fork the mainnet, so we need to point it
       // to the contract artifacts.
       hardhat: process.env.FORKING_URL ? ["./external/mainnet"] : [],
-      goerli: ["./external/goerli"],
-      sepolia: ["./external/sepolia"],
```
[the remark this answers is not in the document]
vzotova: I've noticed only now that I haven't finished with this PR, sorry.
Done! ✔️

manumonti (dismissed):
Bye bye, legacy staking 👋

Please, lmk when the new version of the contract is deployed since it will require some update on subgraphs.

Just a question: why the struct `StakingProviderInfo` in `TokenStaking.sol` is still keeping `nuInTStake` and `keepInTStake`?

vzotova:
> Bye bye, legacy staking 👋
> 
> Please, lmk when the new version of the contract is deployed since it will require some update on subgraphs.
> 
> Just a question: why the struct `StakingProviderInfo` in `TokenStaking.sol` is still keeping `nuInTStake` and `keepInTStake`?

otherwise data layout will be corrupted and other values would not be get correctly. knowing that those slots are 0 we can theoretically reuse them for something new but I'd avoid it too

theref (dismissed):
i love it 🪓

manumonti (approved):

contracts/staking/TokenStaking.sol
```diff
@@ -941,25 +822,14 @@ contract TokenStaking is Initializable, IStaking, Checkpoints {
                 .authorized;
     }
 
-    /// @notice Returns staked amount of T, Keep and Nu for the specified
-    ///         staking provider.
-    /// @dev    All values are in T denomination
-    function stakes(address stakingProvider)
```
cygnusv, line 947: I just realized we will be breaking stuff that consumes Threshold staking information by removing this method. For example, here's the dashboard unpacking the result from `stakes()`:
https://github.com/threshold-network/token-dashboard/blob/8264be8e00286f9d51674fd07ca67030a40ed168/src/threshold-ts/staking/index.ts#L220

What do you think about keeping this method, but marking it as deprecated? Still, I'd keep the `stakeAmount()` method. In the meantime, we can identify stuff that can be updated or removed in the dashboard after legacy support was dropped, which I think it's probably a lot.

docs/rfc-1-staking-contract.adoc
```diff
@@ -369,27 +342,21 @@ each affected application.
 
 Returns the authorized stake amount of the staking provider for the application.
 
-==== `stakes(address stakingProvider) external view returns (uint96 tStake, uint96 keepInTStake, uint96 nuInTStake)`
```
cygnusv, line 372: See my comment above about leaving this method but marked as deprecated

docs/rfc-1-staking-contract.adoc
```diff
@@ -403,21 +370,9 @@ Returns length of application array
...
-* 0 T if KEEP stake type specified i.e. min = 40 T max - (10 T) = 30 T
-* 10 T if NU stake type specified i.e. min = 40 T max - (10 T) = 30 T
-* 0 T if T stake type specified i.e. min = 40 T max = 40 T
+==== `getMaxAuthorization(address stakingProvider) external view returns (uint96)`
 
-In other words, the minimum stake amount for
-the specified stake type is the minimum amount of stake of the given type needed
-to satisfy the maximum application authorization given the staked amounts of the
-T stake type for that staking provider.
+Returns the maximum application authorization
```
cygnusv, line 375:
```suggestion
Returns the maximum application authorization.
```

contracts/staking/TokenStaking.sol
vzotova: ✔️

docs/rfc-1-staking-contract.adoc
vzotova: ✔️

docs/rfc-1-staking-contract.adoc
vzotova: ✔️

PR #161, opened 2023-11-29
TokenStaking: method to increase authorization to max

vzotova:
Based on #159 

This new method will be useful for new stakers especially when `TokenStaking` will have more than 2 approved applications

### DIFF
--- a/contracts/staking/TokenStaking.sol
+++ b/contracts/staking/TokenStaking.sol
@@ -28,12 +28,8 @@ import "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
 import "@openzeppelin/contracts-upgradeable/utils/AddressUpgradeable.sol";
 
 /// @notice TokenStaking is the main staking contract of the Threshold Network.
-///         Apart from the basic usage of enabling T stakes, it also acts as a
-///         sort of "meta-staking" contract, accepting existing legacy NU/KEEP
-///         stakes. Additionally, it serves as application manager for the apps
-///         that run on the Threshold Network. Note that legacy NU/KEEP staking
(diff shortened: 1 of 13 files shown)


manumonti (dismissed):
LGTM! 👌

theref (dismissed):

contracts/staking/TokenStaking.sol
```diff
@@ -369,18 +361,64 @@ contract TokenStaking is Initializable, IStaking, Checkpoints {
...
+            uint96 amount = getAvailableToAuthorize(
+                stakingProvider,
+                application
+            );
+
+            // new application
+            if (amount == tStake) {
+                stakingProviderStruct.authorizedApplications.push(application);
+                authorizedApplications++;
+            }
```
cygnusv, line 405: Does it make sense to put this if block inside the `if (amount > 0)` block below? This can remove the need to make this check when amount == 0, which in any case wouldn't imply pushing a new app.

cygnusv (approved):
LGTM, just a minor comment

contracts/staking/TokenStaking.sol
vzotova: I'm not fan of nested if(s) and for(s) so basically here I'm trying to make it easier to read in expense of one additional check. If you think it's not great approach then I can change it

cygnusv: It's not a strong opinion. If you prefer it this way, I'm fine with that 👍

mswilkison:
Closing as obsolete following the TIP-100 changes merged in #171, which removed new staking and authorization increases. The proposed maximum-authorization convenience method belongs to the earlier staking flow. Any future need for this behavior should be scoped against the current contracts.

PR #164, opened 2024-02-21
Update actions to versions running on Node20

michalinacienciala:
As per
https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/, GitHub has started a deprecation process for the GitHub Actions that run on Node16. We're updating Actions that use this version of Node to newer versions, running on Node20.

### DIFF
--- a/.github/workflows/contracts-docs.yml
+++ b/.github/workflows/contracts-docs.yml
@@ -16,9 +16,9 @@ jobs:
     outputs:
       path-filter: ${{ steps.filter.outputs.path-filter }}
     steps:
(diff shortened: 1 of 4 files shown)


cygnusv (approved):

mswilkison:
Closing as superseded by #182. Its remaining unique changes were the checkout and paths-filter updates in contracts-docs.yml; those are now carried forward in commit 81d83fd using the same versions as #182's other workflows. These CI updates stay in #182, while #185 remains focused on verification of active contracts.

PR #166, opened 2024-03-23
Adds tracking of overall authorized amount for each application

vzotova:
Based on #159 

At some moment in the future we want to have contract that will calculate reward per app on chain. For this purpose such contract will need to know overall amount authorized for each application. This PR adds tracking of those values and also one-time function to initialize variables. It's not feasible to calculate current authorized value in contract because we don't have list of all staking providers on chain. `updateAuthorizedOverall` allows to add or subtract values from current `authorizedOverall`, it needed in case before upgrading will be any change in authorization amount. Suggested upgrading sequence: calculate authorized amounts right before upgrading, upgrade + `updateAuthorizedOverall`, in case there is any tx that changes authorization between two previous steps  - execute `updateAuthorizedOverall` again.


theref (approved):

contracts/staking/TokenStaking.sol
```diff
@@ -328,6 +329,36 @@ contract TokenStaking is Initializable, IStaking, Checkpoints {
...
+            "Wrong input parameters"
+        );
+        for (uint256 i = 0; i < applications.length; i++) {
+            address application = applications[i];
+            ApplicationInfo storage applicationStruct = applicationInfo[
+                application
+            ];
+            int96 authorizedOverall = authorizedOverallValues[i];
+            if (authorizedOverall >= 0) {
+                applicationStruct.authorizedOverall += uint96(
```
cygnusv, line 351: Why not directly setting `applicationStruct.authorizedOverall`, rather than using `int96` increments/decrements?

vzotova: because with just setting value we always can face frontrun tx that will change authorization and now we have to execute one more tx. With addition or subtraction - maximum we need two txs even if one/many will be executed right before upgrade

cygnusv (approved):

mswilkison:
Closing this proposal as obsolete following the TIP-100 changes in #171. Aggregate per-application authorization tracking was intended to support the earlier on-chain staking rewards model; TIP-100 moved beta operator compensation to direct DAO payments and retired the relevant staking flows. The related StableYield proposal in #167 is being retired as well.

PR #172, opened 2025-08-11
Pre-unlock updates

vzotova:
Remove `opt-out` code, return back `requestAuthDecrease`


derekpierre (approved):
🎸

cygnusv (commented):
Approved! Sorry for the late review, but I was (technically still am!) on PTO 😂 
Just to give more context: these changes are a mere continuation of the plan devised in TIP-100. Back at the day, we agreed that when the transition time was coming to an end, we would remove the special authorization methods introduced by #171 and prepare a simplified unlock. This is basically just that.

PR #175, opened 2026-02-10
Touchups

vzotova:


contracts/staking/TokenStaking.sol
```diff
@@ -488,6 +494,36 @@ contract TokenStaking is Initializable, IStaking, Checkpoints {
         );
     }
 
+    /// Migration
+    function migrateAndRelease(address stakingProvider, uint96 amount) external override {
+        ApplicationInfo storage applicationStruct = applicationInfo[msg.sender];
+        require(
+            applicationStruct.status == ApplicationStatus.APPROVED,
+            "Application is not approved"
+        );
```
cygnusv, line 510: Can we remove this check? Given that TACoApp is trusted in this context, it seems a bit redundant.

contracts/staking/TokenStaking.sol
```diff
@@ -488,6 +494,36 @@ contract TokenStaking is Initializable, IStaking, Checkpoints {
...
+
+        require(!skipApplication(msg.sender), "Only TACo app can call this method");
```
cygnusv:
```suggestion
        require(msg.sender == TACO_APPLICATION, "Only TACo app can call this method");
```
I know it's the same but we can be a bit more explicit.

contracts/staking/TokenStaking.sol
```diff
@@ -872,7 +911,7 @@ contract TokenStaking is Initializable, IStaking, Checkpoints {
     // slither-disable-next-line dead-code
     function skipApplication(address application)
         internal
-        pure
+        view
```
cygnusv, line 889: Why is it `view` now? Just curious

contracts/staking/TokenStaking.sol
```diff
@@ -304,6 +303,13 @@ contract TokenStaking is Initializable, IStaking, Checkpoints {
             cleanAuthorizedApplications(stakingProviderStruct, 1);
         }
 
+        // Unstake
+        stakingProviderStruct.tStake -= authorization.deauthorizing;
+        decreaseStakeCheckpoint(stakingProvider, authorization.deauthorizing);
+        emit Unstaked(stakingProvider, authorization.deauthorizing);
+        token.safeTransfer(stakingProviderStruct.owner, authorization.deauthorizing);
+
+        authorization.deauthorizing = 0;
```
cygnusv, line 315: What's the goal of these lines?

cygnusv (commented):
Looks good in general! just some comments

contracts/staking/TokenStaking.sol
vzotova: `skipApp` helps for testing

contracts/staking/TokenStaking.sol
vzotova: I was trying to get same logic as in another methods, and that part is already tested

contracts/staking/TokenStaking.sol
vzotova: just because of the tests, there is override that needs to be `view`

contracts/staking/TokenStaking.sol
vzotova: unstake the rest, everything over minimum

contracts/staking/TokenStaking.sol
cygnusv: ok, fair enough!

contracts/staking/TokenStaking.sol
cygnusv: I see. My concern was cost for a migration of dozens of stakers, but apparently since we do this in a batch TX, only the first access to `applicationStruct.status` would get the full SLOAD cost, and the rest will only be 100 gas.

contracts/staking/TokenStaking.sol
```diff
@@ -495,23 +498,39 @@ contract TokenStaking is Initializable, IStaking, Checkpoints {
...
             .authorizations[msg.sender];
+
+        // stakeless
+        if (authorization.authorized == 0) {
+            return true;
+        }
+
         require(authorization.authorized >= amount, "Not enough authorization");
 
-        decreaseStakeCheckpoint(stakingProvider, stakingProviderStruct.tStake);
```
cygnusv: Why was this line removed? In all cases, `tStake` tokens are leaving the contract, so this line should stay.

vzotova: my bad, missed it through rearranging

vzotova: ✔️

contracts/staking/TokenStaking.sol
```diff
@@ -488,6 +497,59 @@ contract TokenStaking is Initializable, IStaking, Checkpoints {
...
+        StakingProviderInfo storage stakingProviderStruct = stakingProviders[
+            stakingProvider
+        ];
+        require(
+            stakingProviderStruct.owner != address(0),
+            "Wrong staking provider"
+        );
+        uint96 toUnstake = stakingProviderStruct.tStake;
+        stakingProviderStruct.tStake = 0;
+        decreaseStakeCheckpoint(stakingProvider, 0);
```
cygnusv:
```suggestion
        decreaseStakeCheckpoint(stakingProvider, toUnstake);
```
The parameter here is the amount to decrease, so we want here to decrease all.

contracts/staking/TokenStaking.sol
```diff
@@ -488,6 +497,59 @@ contract TokenStaking is Initializable, IStaking, Checkpoints {
...
+        if (authorization.authorized == 0) {
+            stakeless = true;
+        } else {
+            require(
+                authorization.authorized >= amount,
+                "Not enough authorization"
+            );
+            toUnstake -= amount;
+            authorization.authorized = 0;
+            if (amount > 0) {
```
cygnusv, line 542: If we make `amount` fixed, then this condition will always be true.

contracts/staking/TokenStaking.sol
```diff
@@ -488,6 +497,59 @@ contract TokenStaking is Initializable, IStaking, Checkpoints {
...
+                token.safeTransfer(msg.sender, amount);
```
cygnusv, line 543: I think that, technically, this amount is also `Unstaked`, at least from the Threshold Network perspective, so there should be an `Unstaked` event.

test/staking/TokenStaking.test.js
```diff
@@ -1182,6 +1166,296 @@ describe("TokenStaking", () => {
...
+            application1Mock.address
+          )
+        ).to.equal(Zero)
+      })
+
+      it("should emit Unstaked", async () => {
+        await expect(tx)
+          .to.emit(tokenStaking, "Unstaked")
+          .withArgs(stakingProvider.address, amount.sub(amountToTransfer))
+      })
```
cygnusv, line 1340: We need to add voting weight checks here as well. It should go to 0.

contracts/staking/TokenStaking.sol
```diff
@@ -488,6 +497,59 @@ contract TokenStaking is Initializable, IStaking, Checkpoints {
         );
     }
 
+    /// Migration
+    function migrateAndRelease(address stakingProvider, uint96 amount)
```
cygnusv, line 501: I'm starting to think we should fix the amount to migrate as constant/immutable, because that reduces trust in the app implementation and makes the review process . Also it's a tiny gas reduction.

test/staking/TokenStaking.test.js
vzotova: ✔️

contracts/staking/TokenStaking.sol
vzotova: ✔️

contracts/staking/TokenStaking.sol
vzotova: ✔️

mswilkison (approved):

cygnusv (approved):

PR #176, opened 2026-04-07
Support safe Sepolia staking upgrades and operator setup

lionakhnazarov:
Sepolia operators need native T staking and authorization for Random Beacon and Wallet Registry. This adds that workflow through a dedicated `SepoliaTokenStaking` implementation, with proxy deployment, upgrade, funding, keystore, and operator registration tooling.

- Both upgrade entrypoints select the guarded Sepolia implementation. It exposes native staking without test-fixture mutators, emits native-T `Staked` events for indexers, keeps every approved application's authorization backed during decreases and withdrawals, and disables legacy TACo migration on Sepolia. The base contract retains TACo eligibility and rejects deprecated applications before changing authorization.
- Mainnet and Sepolia deployment reruns reuse their recorded proxy. A record pointing to missing code stops deployment for reconciliation. Direct local deployments retain hardhat-deploy's normal artifact/constructor-change handling.
- Sepolia deployment and upgrade resolve artifacts through hardhat-deploy, including `external.contracts`. The package includes matching compiler build information, which the consuming project's OpenZeppelin plugin validates in its own cache format before deploying. Missing or mismatched build information stops the operation before a transaction; existing consumer validation entries remain intact.
- The shipped Sepolia OpenZeppelin manifest now matches the recorded proxy, its current admin, and the historical `ExtendedTokenStaking` layout, including `skipList`. Its implementation bytecode was rebuilt from the deployment commit and matched against Sepolia. The manifest is included in npm distributions; [upgrade-history documentation](https://github.com/threshold-network/solidity-contracts/blob/3620f4dab75fe7ce94a7db2ad51e954abc2e8b71/docs/sepolia-staking-upgrades.adoc) explains provenance and consumer/historical-admin requirements.
- Upgrades preserve deployment metadata and save the implementation/ABI through `deployments.save`, so later operations and exports include `stake()`.
- Credential generation creates private files exclusively, preserves existing wallets on retries, and allows a fresh attempt after preliminary generation fails. Both singular `.env.operator-*` and plural `.env.operators-*` credential files are ignored.
- Funding rejects missing, empty, or unreadable explicit configuration paths before transferring tokens; the default file is used only when no argument is supplied. All operator setup commands load shared, selected, and generated configuration from explicit filesystem paths, preventing same-named files on `PATH` from selecting another provider while preserving each command's working-directory behavior.
- `setup-multiple-operators.sh` uses the same normalized decimal wei amount for the ETH balance check and both wallet transfers. Unitless `ETH_PER_OPERATOR` values are interpreted as ether, so `0.05` and `1` fund each wallet with the intended amount.
- Operator setup propagates each failed transaction and reports incomplete registrations with a nonzero exit status. Ambiguous broadcast or receipt failures stop without rebuilding a potentially non-idempotent transaction; operators must reconcile the original submission before retrying.

Validation on the remediation commits:

- `npm run prepack` (including Solidity compilation, TypeScript and artifact/build-info export) and `tsc -p tsconfig.export.json --noEmit` passed.
- Full Hardhat test suite: **443 passing**, with gas reporting disabled locally. This includes 63 added checks covering contract authorization/backing, staking events, storage-compatible upgrades from the committed Sepolia history, proxy reuse, deployment exports, external artifact consumption, key preservation, configuration selection across operator entrypoints, exact ETH funding amounts, failure propagation, and single-attempt transaction submission.
- A real npm tarball was tested in consumers without locally compiled staking artifacts. Deployment, upgrade, missing-metadata and mismatched-artifact checks passed using both this repository's dependency tree and tBTC's installed Hardhat 2.12.5 / OpenZeppelin plugin 1.22.0 dependency tree.
- `hardhat deploy --network hardhat --export /tmp/pr176-followup-export.json` passed.
- The standalone upgrade command rejected the local Hardhat network before deployment, as expected.
- Repository-pinned Slither 0.8.0: **0 findings**; modified-file ESLint, Solhint, Prettier, credential-ignore and diff checks passed. Bash syntax checks passed in the initial remediation.

The broader `npm run format` command still reports 58 pre-existing style errors in three unchanged keystore scripts; the changed JavaScript and shell files pass the applicable lint, formatting, and syntax checks.

No public-network deployment was performed. Coordinate the shared deployment-script changes with #186 and the toolchain refresh in #182 when integrating these PRs.


contracts/test/TokenStakingTestSet.sol
```diff
@@ -261,7 +261,7 @@ contract ExtendedTokenStaking is TokenStaking {
 
     /// @notice Allows the Governance to approve the particular application
     ///         before individual stake authorizers are able to authorize it.
-    function approveApplication(address application) external {
+    function approveApplication(address application) external override {
```
lrsaturnino: **Critical — Missing `onlyGovernance` modifier**

The base `TokenStaking.approveApplication` (line 269-273) requires `onlyGovernance`, but this override drops it. Solidity's override semantics replace the parent entirely — modifiers are NOT inherited. This means any address can call `approveApplication` on the Sepolia proxy and approve arbitrary applications.

Suggestion: either add `onlyGovernance` here:
```solidity
function approveApplication(address application) external override onlyGovernance {
```

Or remove this override entirely and let the base `TokenStaking.approveApplication` (which already has `onlyGovernance`) handle it — this override is an exact copy minus the access control.

package.json
```diff
@@ -61,5 +62,6 @@
     "@openzeppelin/contracts": "~4.5.0",
     "@openzeppelin/contracts-upgradeable": "~4.5.2",
     "@thesis/solidity-contracts": "github:thesis/solidity-contracts#4985bcf"
-  }
+  },
+  "packageManager": "yarn@4.8.1+sha512.bc946f2a022d7a1a38adfc15b36a66a3807a67629789496c3714dd1703d2e6c6b1c69ff9ec3b43141ac7a1dd853b7685638eb0074300386a59c18df351ef8ff6"
```
lrsaturnino: **Critical — Yarn 4 migration breaks the entire toolchain**

Pinning to Yarn 4.8.1 migrates from Yarn 1 classic to Yarn 4 PnP. Without a `.yarnrc.yml` setting `nodeLinker: node-modules`, PnP strict dependency resolution breaks the build:

- `hardhat compile` fails — `@keep-network/hardhat-helpers` cannot resolve `fs-extra` (undeclared transitive dep)
- `hardhat test` fails identically
- `eslint` fails — cannot find `eslint-plugin-no-only-tests`

Diff-testing the base vs head branches confirmed all 3 fail. Suggestion: add `.yarnrc.yml` with `nodeLinker: node-modules` to maintain compatibility, or declare the missing transitive dependencies explicitly.

lrsaturnino (commented):
Two critical findings from multi-model code review (3/3 models confirmed both).

lrsaturnino (dismissed):
Looks good overall — clean contract changes, CEI ordering solid, access controls in place.

One small nit: the operator setup scripts generate keystores and env files (`operator-1-keystore/`, `spv-maintainer-keystore/`, `.env.new-operator`, `.env.operator-*`) that aren't covered by `.gitignore`. Might be worth adding entries to prevent accidental commits.

deploy/07_deploy_token_staking.ts
```diff
@@ -15,7 +15,7 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
 
   // TODO: Consider upgradable deployment also for sepolia.
   let tokenStakingAddress
-  if (hre.network.name == "mainnet") {
+  if (hre.network.name == "mainnet" || hre.network.name == "sepolia") {
```
lrsaturnino: Enabling this branch for sepolia exposes a path mismatch. The body of this block writes TokenStaking.json at the repo root (line 40) via fs.writeFileSync with OZ upgrades.deployProxy, and there is no subsequent deployments.save, so on a clean Sepolia run hardhat-deploy has no record of the proxy and nothing is ever placed under deployments/sepolia/, which is what deploy/54 then looks for. Adding await deployments.save("TokenStaking", tokenStakingDeployment) here, or writing to deployments/${hre.network.name}/TokenStaking.json alongside the root write, should keep the downstream upgrade script working.

deploy/54_upgrade_token_staking_extended.ts
```diff
@@ -0,0 +1,85 @@
...
+    return
+  }
+
+  let proxyAddress: string
+  const existing = await deployments.getOrNull("TokenStaking")
+  if (existing) {
+    proxyAddress = existing.address
+  } else {
+    // 07_deploy_token_staking saves to TokenStaking.json in deployments dir
+    const deploymentPath = `deployments/${hre.network.name}/TokenStaking.json`
```
lrsaturnino, line 32: This fallback reads deployments/${hre.network.name}/TokenStaking.json, but deploy/07's sepolia branch writes the snapshot at the repo root instead and never calls deployments.save, so deployments.getOrNull("TokenStaking") on line 28 also returns null. On a clean Sepolia run that means line 34 evaluates false and the script takes line 35 and silently no-ops rather than upgrading. Aligning the write path in deploy/07, or having this fallback also check the root TokenStaking.json, would unblock the upgrade.

deploy/54_upgrade_token_staking_extended.ts
```diff
@@ -0,0 +1,85 @@
...
+
+/**
+ * Upgrades TokenStaking proxy to ExtendedTokenStaking implementation.
+ * ExtendedTokenStaking adds the stake() function required for native T staking.
+ * The base TokenStaking does not have stake() - it only supports legacy KEEP/NU
+ * migrations. Run this on sepolia (or other testnets) where operators need to
+ * stake T directly.
+ *
+ * From repo root (syncs deployments, upgrades, copies ABI to tbtc-v2):
+ *   bash scripts/upgrade-token-staking-sepolia.sh
```
lrsaturnino: scripts/upgrade-token-staking-sepolia.sh is not in the tree at b524ca4 (only scripts/upgrade-token-staking.ts exists), so operators following this docstring will hit command not found. Either adding the wrapper or replacing this reference with the direct hardhat invocation would avoid the confusion.

scripts/upgrade-token-staking.ts
```diff
@@ -0,0 +1,88 @@
+/**
+ * Upgrades the TokenStaking proxy to ExtendedTokenStaking (adds stake()).
+ * Run: npx hardhat run scripts/upgrade-token-staking.ts --network sepolia
+ * Or from repo root: bash scripts/upgrade-token-staking-sepolia.sh
```
lrsaturnino: Same missing file as referenced in deploy/54 — scripts/upgrade-token-staking-sepolia.sh is not in the repo. It would be worth either adding it or pointing operators directly at npx hardhat run scripts/upgrade-token-staking.ts --network sepolia.

scripts/setup-multiple-operators.sh
```diff
@@ -0,0 +1,201 @@
...
+  local op_key="OP${i}_OPERATOR_KEY"
+  sp_addr=${!sp_addr}
+  sp_key=${!sp_key}
+  op_addr=${!op_addr}
+  op_key=${!op_key}
+  [ -n "$sp_addr" ] && [ -n "$sp_key" ] && [ -n "$op_addr" ] && [ -n "$op_key" ] || return 1
+  echo "--- Operator $i/$N (existing) ---"
+  cast_send_ok $TOKEN_STAKING "increaseAuthorization(address,address,uint96)" \
+    "$sp_addr" $RANDOM_BEACON $AMOUNT_40K \
+    --rpc-url $CHAIN_API_URL --private-key "$sp_key"
```
lrsaturnino: This is the first of seventeen --private-key CLI exposures in this file (lines 124, 127, 129, 131, 133, 135, 166, 170, 172, 176, 180, 183, 186, 188, 190, 192, 194), and each one leaks its key into ps aux and the caller's shell history. Commit 42ca14f moved run-new-operator-setup.sh and fund-new-operator.sh to ETH_PRIVATE_KEY="$key" cast send ... for exactly this reason. Since this file is new in the PR and drives the batch flow, it would be good to finish that same hardening here.

lrsaturnino (commented):
Some more findings for evaluation.

lrsaturnino (commented):
On `ExtendedTokenStaking` in `contracts/test/TokenStakingTestSet.sol` — now that `approveApplication` and `increaseAuthorization` live on the base `TokenStaking`, the overrides at lines 227–268 and 280–294 are duplicates. The `increaseAuthorization` override also doesn't emit `AuthorizationIncreased` or guard `application != address(0)`, while the base does both.

Since `deploy/54` makes `ExtendedTokenStaking` the live Sepolia implementation, Sepolia will silently stop emitting `AuthorizationIncreased` (no test asserts it, so CI won't flag it). Dropping both overrides should be enough — the derived contract picks up the canonical behaviour from the base.

lrsaturnino:
`.openzeppelin/unknown-11155111.json` (the OZ Sepolia manifest) doesn't seem to be tracked, and `.gitignore:21` matches the `.openzeppelin/unknown-*.json` pattern that covers it. That could leave a fresh clone tripping on `Manifest not found` the next time someone runs the Sepolia upgrade, and the storage-layout history wouldn't carry forward across upgrades.

Probably a quick follow-up — commit the manifest (from whoever ran `deployProxy` originally) plus a `!.openzeppelin/unknown-11155111.json` carve-out after `.gitignore:21`, mirroring how `mainnet.json` is already tracked.

mswilkison:
Addressed all seven review findings in 0fab89a:

| Review finding | Remediation and regression evidence |
| --- | --- |
| Public deployment selected a test fixture | Both upgrade entrypoints share the Sepolia-only upgrade function and select `SepoliaTokenStaking`; ABI and authorization tests exclude test mutators and enforce governance/authorizer controls. |
| Reruns replaced the Sepolia proxy | Existing mainnet/Sepolia proxy records are reused, missing code fails before replacement, and tests verify address/state preservation. Local direct deployment refresh is preserved. |
| Deprecated application authorizations were inconsistent | The base entrypoint applies its eligibility guard. Sepolia supports the complete approved-application lifecycle, including multi-application withdrawal backing and disabled legacy migration. |
| Setup overwrote staking-provider keys | Exclusive private env/keystore creation preserves existing credentials. Tests compare original file hashes across retries, reject symlinks, and cover preliminary directory failures. |
| Existing-operator setup hid transaction failures | Every transaction explicitly returns failure; partial registrations make the script exit nonzero. Each of the six transaction positions is covered. |
| Upgraded ABI was missing from exports | `deployments.save` updates the registry and disk together, preserving metadata. Tests confirm exported `stake()`. |
| Ambiguous broadcasts could duplicate transfers | Submission and receipt errors stop without reconstructing the send. Mocked cases confirm exactly one send for already-known, nonce, missing-hash, and receipt failures. |

Validation: 410 tests passed (30 new), compilation, packaging, local deployment/export, repository-pinned Slither 0.8.0 (0 findings), modified-file lint/format, and Bash syntax. Storage upgrade checks cover both the base contract and previous fixture layout.

Independent candidate review also identified local redeployment and failed-generation retry regressions; both are corrected and included in the tests. These are source fixes; no Sepolia or mainnet deployment was performed.

GitHub CI on 0fab89a is green: build/tests, deployment dry run, Slither, formatting workflow, and documentation preview all passed.

mswilkison:
Fixed the three follow-up findings in [a8fe18c](https://github.com/threshold-network/solidity-contracts/commit/a8fe18c73f15168dd183e1692b1a3255b10c7374).

1. **External artifacts and upgrade validation:** Both Sepolia entrypoints resolve factories through hardhat-deploy's artifact registry. The package now includes matching compiler build information, and the consuming project's OpenZeppelin plugin regenerates its own validation entries. Existing entries remain intact, and the normal deployment/storage-layout checks remain enabled. Missing metadata or a bytecode/ABI mismatch is rejected before a transaction.
2. **Staking events:** `SepoliaTokenStaking.stake` emits the inherited native-T `Staked` event with the owner, provider, beneficiary, authorizer and amount. The regression test checks every field, exactly one event, the stake balance, and the matching withdrawal.
3. **Credential filenames:** `.gitignore` now covers `.env.operators-*`, including the documented `.env.operators-3` file and nested equivalents.

Changed files: `deploy/07_deploy_token_staking.ts`, `deploy/54_upgrade_token_staking_extended.ts`, `scripts/staking-artifacts.ts`, `scripts/export-staking-build-info.js`, `package.json`, `contracts/staking/SepoliaTokenStaking.sol`, `.gitignore`, `test/deployment/ExternalStakingDeployment.test.js`, and `test/staking/SepoliaTokenStaking.test.js`.

Verification, in order:

- Build/type checks: `npm run prepack` and `tsc -p tsconfig.export.json --noEmit` passed.
- Focused regression and compatibility checks: an actual npm tarball deploys and upgrades in an `external.contracts` consumer with no locally compiled staking artifacts. The proxy address, stored configuration, deployment metadata, exported ABI and pre-existing validation entries are preserved. Missing-metadata and mismatched-artifact cases reject before sending a transaction. These checks passed with this repository's Hardhat 2.10.2 / OpenZeppelin plugin 1.14.0 dependencies and tBTC's installed Hardhat 2.12.5 / plugin 1.22.0 dependencies.
- Full suite: **414 passing**, using matched compiled artifacts and gas reporting disabled locally:
  ```js
  const hre = require("hardhat")
  hre.config.gasReporter = { enabled: false }
  await hre.run("test", { noCompile: true })
  ```
  Run through Node 18 with `ts-node/register`; test failures propagate as a nonzero exit status.
- Local deployment: `hardhat deploy --network hardhat --export /tmp/pr176-followup-export.json` passed.
- Static checks: repository-pinned `slither . --ignore-compile` (0.8.0) reported **0 findings**. Modified-file ESLint, Solhint, Prettier, credential-ignore checks and `git diff --check` passed.
- An independent read-only review found no surviving bypasses or regressions in this patch.

No public-network deployment or live subgraph validation was performed. The downstream checks use tBTC's dependency versions in isolated local consumers; they do not deploy the entire tBTC system. GitHub CI results will be recorded separately.

mswilkison:
Fixed the CI installation failure in [a7ba9ac](https://github.com/threshold-network/solidity-contracts/commit/a7ba9ac627de65d2318ab93c00acba412b73014f).

The [failed run](https://github.com/threshold-network/solidity-contracts/actions/runs/34282551444/job/102250512413) stopped while installing dependencies: concurrent Yarn Classic bootstraps for Git dependencies shared the runner's cache and hit an `ENOENT` for `lodash/fp/valuesIn.js`.

The Solidity, format, npm and reusable docs workflows now pass `YARN_CACHE_FOLDER=.yarn/cache`. Yarn Classic resolves that relative path inside each dependency's temporary checkout, so parallel bootstraps have separate caches. Yarn 4 retains its existing global cache behavior. Dependencies and the lockfile are unchanged.

Verified that two concurrent Yarn Classic 1.22.22 processes resolve different cache directories, and that all four workflows parse correctly. All checks passed on this commit: [Solidity](https://github.com/threshold-network/solidity-contracts/actions/runs/34283012530) (build/tests, deployment dry run and Slither), [format](https://github.com/threshold-network/solidity-contracts/actions/runs/34283012518), and [documentation preview](https://github.com/threshold-network/solidity-contracts/actions/runs/34283012806).

mswilkison:
Addressed both new findings in 3620f4dab75fe7ce94a7db2ad51e954abc2e8b71.

- Restored the recorded Sepolia proxy's matching OpenZeppelin implementation history and ProxyAdmin. The complete historical layout includes `skipList`; all 45 previous implementation entries and 33 previous proxy records remain intact. The manifest is now included in the npm package.
- Funding now rejects a missing, empty, directory, or unreadable explicit configuration path before transferring tokens. Defaults apply only when no argument is supplied. Explicit and default loads are anchored to the checked filesystem paths, so a same-named file on `PATH` cannot select a different provider.

The manifest recovery is based on read-only Sepolia evidence at block **11663843**. Recompiling historical commit `ad7c06e23672ca7b787f6bb882d519e0e5851dc3` with Solidity `0.8.9+commit.e5eed63a` matches the deployed implementation byte-for-byte, including metadata and all seven immutable token references. A separate reviewer independently reproduced the compiler/layout match and corroborated the proxy, implementation, and admin through public RPC. See the [provenance and upgrade instructions](https://github.com/threshold-network/solidity-contracts/blob/3620f4dab75fe7ce94a7db2ad51e954abc2e8b71/docs/sepolia-staking-upgrades.adoc).

Verification:

- **Syntax and static checks:** `bash -n scripts/fund-new-operator.sh`, ESLint for the three changed JavaScript test files, Prettier for the changed files covered by repository configuration, and `git diff --check` passed.
- **Regression checks:** before the patch, upgrading the recorded snapshot failed with “Deployment at address ... is not registered,” and invalid explicit funding paths selected the default wallet. The three recorded-history cases and seven funding cases now pass. Both upgrade entrypoints preserve the proxy, staking balances, roles, authorizations, metadata, and stored `skipList` values; deleting `skipList` still fails storage validation before a transaction. Relative, absolute, default, and environment funding modes still select the intended provider, including with shadow files on `PATH`.
- **Integration:** the complete Hardhat suite passed: **424 tests** with local gas reporting disabled. This runs `npm run prepack`, creates an actual npm tarball, verifies the shipped manifest, and exercises external-artifact consumers. The operator subset passed all **26 tests**. All upgrade transactions were confined to in-process Hardhat networks.

The repository-wide `npm run format` still reports **58 pre-existing style errors** in the unchanged `create-operator-keystore.js`, `create-spv-maintainer-keystore.js`, and `get-operator-key.js` scripts; each is byte-identical to the prior PR head. The OpenZeppelin manifest retains its generated JSON formatting, as configured by the repository's ignore rules.

Compatibility: OpenZeppelin 1.14 has one admin per manifest. The active entry now matches the currently recorded Sepolia proxy; older proxies with the historical admin require their matching historical checkout/manifest. External consumers retain their own manifest and must incorporate the recovered history as documented. No validation bypass or automatic manifest overwrite was added.

GitHub CI passed for head `3620f4dab75fe7ce94a7db2ad51e954abc2e8b71`: [build/tests, deployment dry run, and Slither](https://github.com/threshold-network/solidity-contracts/actions/runs/34286419718), [format workflow](https://github.com/threshold-network/solidity-contracts/actions/runs/34286419769), and [documentation preview](https://github.com/threshold-network/solidity-contracts/actions/runs/34286419897). The CI test log confirms **424 passing**; Slither reports **0 results**. Public deployment and release-only jobs were skipped as expected. No public-network deployment or transfer was performed.

mswilkison:
Addressed both operator-tooling findings in 594bf497b40a8ad34ac061399e7997c5a120cdcf.

- `setup-multiple-operators.sh` now anchors relative `OPERATORS_CONFIG` paths before checking and sourcing them, and uses explicit filesystem paths for shared `.env` and generated `.env.operator-*` files. `run-new-operator-setup.sh` does the same for both of its configuration loads. Each command retains its original working-directory behavior. Usage examples in these scripts and `fund-new-operator.sh` also use explicit paths.
- Both ETH transfers now send the quoted `_eth_per_operator_wei_dec` value used for preflight. Unitless `0.05` sends **50000000000000000 wei** to each wallet; unitless `1` sends **1000000000000000000 wei**. The existing amount parser and balance requirements are preserved.

Added 19 regression cases in `test/scripts/OperatorConfiguration.test.js`, using complete script execution with local transaction/generation stubs and dummy credentials. Before the patch, the harness reproduced loading shadow files from `PATH` and forwarding raw funding values instead of the checked wei amount. Those triggers now pass. Controls cover default, absolute, relative, parent-relative, space-containing and leading-dash paths; automatic existing mode; generated-file absence; invocation-directory and environment-only setup; default/suffixed and unitless funding amounts; hex conversion output; multiple operators; and stopping before generation or transfers for invalid amounts or insufficient ETH.

Verification, in order:

- `bash -n` on each of the three changed shell scripts, ESLint for the new test file, Prettier for all four changed files, and `git diff --check`: passed.
- `node node_modules/mocha/bin/_mocha test/scripts/OperatorConfiguration.test.js test/scripts/OperatorSetup.test.js`: **45 passing** (19 new, 26 existing).
- Full Hardhat suite through `hre.run("test", { noCompile: true })` with local gas reporting disabled: **443 passing**, including the actual package/prepack and external deployment tests.
- Independent candidate review: no surviving source-backed issue found. The reviewer also passed the 45 operator checks, Bash 3.2 checks for unusual filenames and symlinks, and offline Cast amount conversions.

The previously reported repository-wide formatting errors in unchanged keystore scripts are outside this patch; changed-file checks pass. No public-chain deployment or transfer was performed.

GitHub CI passed for head `594bf497b40a8ad34ac061399e7997c5a120cdcf`: [Solidity checks](https://github.com/threshold-network/solidity-contracts/actions/runs/34290000436), [format workflow](https://github.com/threshold-network/solidity-contracts/actions/runs/34290000574), and [documentation preview](https://github.com/threshold-network/solidity-contracts/actions/runs/34290001045). The CI test log confirms **443 passing**; Slither reports **0 results**. The first change-detector attempt encountered GitHub's temporary “diff unavailable due to heavy server load” error; retrying only that job passed without a code change.

PR #179, opened 2026-05-05
fix: address review findings from #176

piotr-roslaniec:
## Summary

Follow-up to #176. Addresses all P1/P2/P3 findings from the multi-agent review.

**P1 fixes:**
- `deploy/54_upgrade_token_staking_extended.ts`: add `kind: "transparent"` to `upgradeProxy` call; removes reliance on OZ manifest for proxy-type detection
- `scripts/setup-new-staking-provider.js`: `chmod 0600` env file after write (private keys were world-readable under default umask)
- `scripts/setup-multiple-operators.sh`: replace `cast wallet address --private-key "$KEY"` with `ETH_PRIVATE_KEY="$KEY" cast wallet address` to avoid key exposure in process table
- `deployments/mainnet/TokenStaking.json`: restore correct 6-param constructor ABI (PR had accidentally replaced it with a 2-param variant, breaking ABI correctness for the deployed mainnet contract)

**P2 fixes:**
- Extract shared `cast_send_ok` with retry/backoff into `scripts/lib/cast-helpers.sh`; both `run-new-operator-setup.sh` and `setup-multiple-operators.sh` source it. Previously `run-new-operator-setup.sh` had no retry logic.
- `scripts/get-operator-key.js`: `--list` mode now reads the unencrypted `address` field from keystore JSON directly, instead of attempting decryption with an empty password
- `scripts/fund-new-operator.sh`: remove hardcoded fallback T address (`0x60185Ef...`); fail loudly with an error if `T.json` is missing; quote all `cast send` variables
- `deploy/07_deploy_token_staking.ts`: remove dual-write to root `TokenStaking.json` (only write to `deployments/<network>/`); delete the previously committed root artifact

**Also fixed (pre-existing hook findings triggered by touched files):**
- Replace sync `fs.*Sync` calls with `fs.promises` in deploy scripts
- Wrap `JSON.parse` in try/catch where flagged
- Replace `==` with `===` in network name checks

## Test plan

- [ ] CI passes on this branch
- [ ] `cast wallet address` works with `ETH_PRIVATE_KEY` env var (Foundry supports it)
- [ ] `node scripts/get-operator-key.js --list` lists addresses without password errors
- [ ] `deployments/mainnet/TokenStaking.json` constructor shows 6 params matching the live contract


lionakhnazarov (approved):
setup-multiple-operators.sh should derive addresses with cast wallet address <pk> or --private-key, not ETH_PRIVATE_KEY=… cast wallet address (Foundry doesn’t wire that env for wallet address). - fixed in following commit

except this - LGTM

